PR #14, opened 2015-10-21
volmaster: move volume iteration code to a generic function

erikh:


erikh:
LGTM

PR #15, opened 2015-10-21
Go away go-etcd

erikh:
This removes the initially used, and now deprecated `go-etcd`. It now uses `github.com/coreos/etcd/client` which is the canonical go client.


erikh:
LGTM

PR #20, opened 2015-10-22
Fix ceph ansible

erikh:
This provides a number of fixes to the build to make it more reliable. In particular, the RPM that ceph uses to install its repository data relies on a URL for the GPG key which is not always available. We try to provide multiple solutions here to resolve the problem.


erikh:
LGTM

PR #23, opened 2015-10-25
new volsupervisor binary replaces scheduled tasks in volmaster and makes volmaster multi-master.

### LINKED ISSUE #16: ephemeral volumes do not get cleaned up on host disappearance
Right now, if the host disappears and docker loses state of the volume, we leak a volume that was intended to be deleted upon container/volume teardown inside of docker.

To resolve this, I propose we implement a scheduled jobs daemon that handles this and snapshots, and of course, can be used for more later. This `volsupervisor` could live on one host with a connection to the volmaster and etcd, and can be asynchronously managed with the rest of the cluster due to its tertiary purpose.

erikh:
This partially resolves #16 and also means that the volmaster can now be used on multiple hosts at once (and is tested that way).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 start: download-docker install-ansible
 	vagrant up
-	make build run
+	make build
 
 stop:
 	vagrant destroy -f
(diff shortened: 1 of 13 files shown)


erikh:
LGTM

PR #26, opened 2015-10-27
Rate limit io

erikh:
This implements rate limiting. It comes in three distinct parts:

* config -- parameters, json configuration
* cephdriver -- code to gather device major/minor numbers and report it (and other critical data) with the mount.
* volplugin -- code to implement the cgroups file writes to the blkio controller on mount.

the blkio directives we're utilizing automatically vaporize on a umount, which is what we do any time the container is stopped. Therefore, applying them each time becomes idempotent, even with multiple hosts and mounts going on at the same time.

The tests verify the files are written to with magic numbers; they are high enough in value that they allow the fsck to proceed. :)

the major/minor device number code is pretty linux specific, and probably bad.


erikh:
LGTM

PR #29, opened 2015-10-31
Abstract ceph away

erikh:
This provides a new driver framework which the Ceph implementation conforms to. In a second set of patches the abstraction will be exposed to end-users, but for now we just use the ceph driver as we always have with the updated API.

/cc @jainvipin @unclejack @mapuri


mapuri:
LGTM. I like how the storgae driver interface use driver options as a generic way to support multiple backends.

I guess once we add support for a second backend it will help bring out the changes, if any, in the frontend (libnetwork facing) that consumes this API.

erikh:
Yeah, I'm working on LVM support, but it's probably a week or two out. At that point things will need to change dramatically on the UI end of things.

erikh:
ok, i'm going to merge this. Thanks @mapuri for the review.

PR #35, opened 2015-11-09
Contivctl changes

erikh:
see commit list, these are for integrating into github.com/contiv/contivctl which is a unified binary for volplugin and netplugin.


erikh:
LGTM

PR #38, opened 2015-11-11
'Use' locking

erikh:
This establishes the concept of 'uses', which are basically mutexes that live in etcd and are controlled by compare and `<action>` operations. We were already doing this for mounts, so I just abstracted that a bit and applied it to create and remove operations as well, and set it up so that they're all mutually exclusive with each other so you can't remove a mounted volume, etc.

There are also some basic tests.

Note that until we have TTL with updates in volmaster/volplugin this is a total disaster area if the host goes down. See #16 for a similar problem.


unclejack:
code LGTM (non-authoritative)

PR #47, opened 2015-11-24
Vagrantfile,ansible: Move to infernalis (ceph 9.2.0)

erikh:


erikh:
Since this is basically an upgrade of the ceph ansible, I am just going to merge this.

mapuri:
@erikh I think we need to do this change in `contiv/build` as well.

Actually we should vendor `contiv/build` in volplugin and use the playbooks there. This way we can keep ansible playbooks and code managed in one repo. Let me know if you would like to do this change or I can take a shot at it.

erikh:
I have some ideas on strategy for contiv/build that we should discuss, 
especially considering it's still locked into being a private 
repository. As for the patch, I'll bring this over as soon as it's 
proven over here.

-Erik

mapuri:
SGTM, thanks @erikh 

Let's discuss the way to use `contiv/build` across project. I already asked @DivyaVavili to help with this on netplugin side. May be we can discuss this today in team meeting.

PR #52, opened 2015-11-27
Makefile: add & use go vet for unit testing

unclejack:


erikh:
does it pass right now?

unclejack:
@erikh Yes, it was passing for me. Please feel free to check. You can also remove `-composites=false` to confirm it doesn't pass when something is wrong.

erikh:
Your word is good enough for me.

PR #57, opened 2015-11-29
storage: driver interface + ceph implementation of ShowMapped

erikh:
@unclejack @mapuri PTAL, this is important to another change I'm working on, but wanted to keep the review set simple.

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -20,6 +20,8 @@ const (
 	mountBase  = "/mnt/ceph"
 )
 
+var spaceSplitRegex = regexp.MustCompile(`\s+`)
+
 // Driver implements a ceph backed storage driver for volplugin.
 //
 // -- Pool naming
@@ -228,7 +230,7 @@ func (c *Driver) ListSnapshots(do storage.DriverOptions) ([]string, error) {
 	lines := strings.Split(string(out), "\n")
 	if len(lines) > 1 {
 		for _, line := range lines[1:] {
-			parts := regexp.MustCompile(`\s+`).Split(line, -1)
+			parts := spaceSplitRegex.Split(line, -1)
 			if len(parts) < 3 {
 				continue
 			}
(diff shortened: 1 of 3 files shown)


storage/driver.go
```diff
@@ -73,4 +73,8 @@ type Driver interface {
 	// ListSnapshots returns an array of snapshot names provided a maximum number
 	// of snapshots to be returned. Any error will be returned.
 	ListSnapshots(DriverOptions) ([]string, error)
+
+	// ShowMapped shows any volumes that belong to volplugin on the host, in
```
mapuri: @erikh 

A quick clarification. Are you planning to use this API for showing the mount information in one of the volcli outputs? I ask as `Mount` struct seems very host specific, do you think for multihost environment this API shall dump some host-info as well (like a host label or hostname)? Or will that be taken care at the volmaster level?

mapuri:
LGTM

storage/driver.go
erikh: No, this is for gathering host resources on volplugin reboot -- so we can better handle host failures.

mapuri: ah...k, gotcha

storage/backend/ceph/ceph.go
```diff
@@ -239,3 +239,38 @@ func (c *Driver) ListSnapshots(do storage.DriverOptions) ([]string, error) {
...
+	}
+
+	// FIXME maybe a new type here would be better than re-using this one.
+	mounts := []*storage.Mount{}
+	for i, line := range strings.Split(string(out), "\n") {
+		if i == 0 {
+			continue
+		}
+
+		parts := regexp.MustCompile(`\s+`).Split(line, -1)
```
unclejack: We should compile this regex only once and have it in the init. This can also be changed afterwards if you prefer that.

erikh: fair enough, I will adjust.

unclejack:
LGTM

Issue #60, opened 2015-11-30
requestVolumeConfig with tenant/volume or tenant/name?


bnoon:
I am just testing this out and I get an error when creating a volume and attempting to mount it into a container.

Here are some logs from volplugin:
```
time="2015-11-30T19:15:48Z" level=debug msg="Dispatching Path with {\"Name\":\"tenant1/foo\"}" 
time="2015-11-30T19:15:48Z" level=info msg="Returning mount path to docker for volume: \"tenant1/foo\"" 
time="2015-11-30T19:15:48Z" level=warning msg="Returning HTTP error handling plugin negotiation: Requesting tenant configuration Status was not 200: was 404: \"\"" 
time="2015-11-30T19:15:48Z" level=debug msg="Dispatching Mount with {\"Name\":\"tenant1/foo\"}" 
time="2015-11-30T19:15:48Z" level=info msg="Mounting volume \"tenant1/foo\"" 
time="2015-11-30T19:15:48Z" level=warning msg="Returning HTTP error handling plugin negotiation: Could not determine tenant configuration Status was not 200: was 404: \"\"" 
```

The logs from volmaster shows that the it get the requests:
```
time="2015-11-30T19:15:48Z" level=debug msg="Dispatching /request with {\"volume\":\"tenant1/foo\",\"tenant\":\"tenant1\"}" 
time="2015-11-30T19:15:48Z" level=debug msg="Dispatching /request with {\"volume\":\"foo\",\"tenant\":\"tenant1\"}" 
```
but the `volume` is different.

should https://github.com/contiv/volplugin/blob/master/volplugin/handlers.go#L132

be `volConfig, err := requestVolumeConfig(master, uc.Tenant, uc.Name)`?

erikh:
Interesting, can you give me the command you ran for `docker volume create` and `docker run` ?

bnoon:
Sure.
```
docker volume create -d volplugin --name=tenant1/foo
docker run -it -v tenant1/foo:/mnt ubuntu bash
```
I had uploaded the canned `systemtests/testdata/intent1.json` file for tenant1.

I am running this against an existing ceph cluster with ceph running in containers.

erikh:
Ok. Can you `rbd ls rbd` for me and see if you get a volume named `tenant1.foo`? Also, if you could upload the output of `etcdctl ls --recursive /volplugin` that would be immensely useful.

Thank you very much for helping me debug this!

bnoon:
[...]

PR #62, opened 2015-12-01
volplugin: fix a bug in handlers.go where the fully qualified path would be passed instead of just the volume name.

erikh:
/cc @unclejack 
also see #60


unclejack:
LGTM

PR #69, opened 2015-12-03
volcli,volmaster,volplugin,volsupervisor: handle errors

unclejack:


volplugin/volplugin.go
```diff
@@ -32,7 +32,9 @@ type VolumeResponse struct {
 // Daemon starts the volplugin service.
 func Daemon(debug bool, master, host string) error {
 	driverPath := path.Join(basePath, "volplugin.sock")
-	os.Remove(driverPath)
+	if err := os.Remove(driverPath); err != nil && !os.IsPermission(err) {
```
erikh: We should probably just do a blanket check for the error here and yield if any. If the permissions are invalid, we still won't be able to create the socket file later.

erikh:
Other comments:

Can you emulate this behavior on `app.Run` failure? https://github.com/contiv/volplugin/blob/master/volcli/volcli.go#L18-L24

unclejack:
@erikh PTAL

erikh:
LGTM

PR #73, opened 2015-12-07
storage: rename ShowMapped to Mounted, gather mount points and other critical data from host.

### LINKED ISSUE #61: Retrieve mount points from /proc/self/mountinfo
This will enable better volplugin recovery on host restart.

erikh:
@unclejack @mapuri @jainvipin PTAL

closes #61

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -242,36 +242,41 @@ func (c *Driver) ListSnapshots(do storage.DriverOptions) ([]string, error) {
 	return names, nil
 }
 
-// ShowMapped describes all the volumes currently mapped on to the host. It
-// does not yield any mount points (yet.)
-func (c *Driver) ShowMapped() ([]*storage.Mount, error) {
-	out, err := exec.Command("rbd", "showmapped").Output()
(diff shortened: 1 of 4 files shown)


mapuri:
LGTM

Just for my understanding, #61 refers to issue of restoring mounts on host restart but I don't see that instrumentation in this PR (unless I missed it), is it coming later? I am curious as I would like to see if it is always ok to just restore state from the host when the statestore might have changed while the host was down.

erikh:
There should be a separate ticket for that. I may have closed it. I will 
create a new one.

This one just concedes that we need this support to continue. I have 
another patch I'm working on to implement the orchestration, I just 
wanted to get this in prematurely so it could be done more 
incrementally.

erikh:
@mapuri I filed #76 to cover the full patch set required, PTAL and comment if you have any concerns about the design; I'd love your feedback.

mapuri:
ok, thanks for clarifying @erikh , i will review #76 and post my comments there.

erikh:
@unclejack PTAL when you are available. I have another patch in queue to file after this one is merged.

Issue #76, opened 2015-12-07
support TTL-backed operations in mount, use locks


erikh:
Use locks and mount operations both need to expire when the host dies, so that other hosts can reschedule the job and get the volumes they need.

Currently we have no support for this; the volumes will linger in etcd until someone manually clears them., and cannot be mounted anywhere else by volplugin. This is unacceptable.

Solution: implement TTL backed publishing of records to etcd. This will allow a failing volplugin to expire gracefully.

Requirements:

* TTL support must be refreshed periodically. Goroutines will accomplish this task trivially. This patch is ready and will be delivered at the end of this project.
* In the case we *restart* volplugin without clearing mounts or restarting the system, that state will be lost, so we must re-establish it. #73 and #57 accomplish this part of the equation by scanning both ceph and existing mounts to establish a unified set of mount points to check against the etcd store. This will allow us to validate that nobody else has attempted to acquire the lock in the meantime.
* We should also have a "maintenance mode" which performs no operations but refreshing etcd and removing mounts (no new mounts). This will allow the mounts to be held still by the containers, and locks can be managed safely between machines. This will be sent in an independent pull request.
* Updating etcd should go through the volmaster, as the volplugin may not have access to it. This patch is ready and will be delivered at the end of this project.
* Additionally, some rewiring of data structures in the use lock system may be required.

erikh:
this was completed some time ago.

PR #78, opened 2015-12-08
config,volmaster,volplugin: first pass at TTL support

erikh:
@mapuri @unclejack @jainvipin 

This implements TTL support or crash-resistant recovery from #76. It implements this by using etcd TTL support to automatically expire records when the host does not report back.

The host may re-acquire the lock on volplugin restart if if the lock has not be re-acquired.

The third case where volplugin is not restarted (therefore the locks are not refreshed) and the mounts are not cleared, this will be settled by another PR which will introduce a "maintenance mode" that people can run to perform host operations without being able to acquire mounts and whatnot. This allows people to keep the locks in place while performing maintenance.

This is a large, critical change so please PTAL (all of you) and review carefully. @shaleman I know you don't work on this project but I'd love to hear your input as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ unit-test-nocoverage:
 unit-test-nocoverage-host: golint-host govet-host
 	HOST_TEST=1 GOGC=1000 godep go test -v ./... -check.v
 
-build: golint
+build: golint govet
 	vagrant ssh mon0 -c 'sudo -i sh -c "cd $(GUESTGOPATH); make run-build"'
 
 run:
--- a/volmaster/daemon.go
+++ b/volmaster/daemon.go
@@ -9,27 +9,31 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/volplugin/config"
+	"github.com/coreos/etcd/client"
 	"github.com/gorilla/mux"
 )
 
 type daemonConfig struct {
-	config *config.TopLevelConfig
+	config   *config.TopLevelConfig
+	mountTTL int
 }
 
 // Daemon initializes the daemon for use.
-func Daemon(config *config.TopLevelConfig, debug bool, listen string) {
-	d := daemonConfig{config}
+func Daemon(config *config.TopLevelConfig, ttl int, debug bool, listen string) {
+	d := daemonConfig{config, ttl}
 	r := mux.NewRouter()
 
 	router := map[string]func(http.ResponseWriter, *http.Request){
-		"/request": d.handleRequest,
-		"/create":  d.handleCreate,
-		"/mount":   d.handleMount,
-		"/unmount": d.handleUnmount,
-		"/remove":  d.handleRemove,
+		"/request":      d.handleRequest,
+		"/create":       d.handleCreate,
+		"/mount":        d.handleMount,
+		"/mount-report": d.handleMountReport,
+		"/unmount":      d.handleUnmount,
+		"/remove":       d.handleRemove,
 	}
 
 	for path, f := range router {
@@ -129,16 +133,30 @@ func (d daemonConfig) handleUnmount(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (d daemonConfig) handleMount(w http.ResponseWriter, r *http.Request) {
+func (d daemonConfig) handleMountWithTTLFlag(w http.ResponseWriter, r *http.Request, exist client.PrevExistType) error {
 	req, err := unmarshalUseConfig(r)
 	if err != nil {
-		httpError(w, "Unmarshalling request", err)
-		return
+		return err
 	}
 
 	req.Reason = "Mount"
 
-	if err := d.config.PublishUse(req); err != nil {
+	if err := d.config.PublishUseWithTTL(req, time.Duration(d.mountTTL)*time.Second, exist); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d daemonConfig) handleMount(w http.ResponseWriter, r *http.Request) {
+	if err := d.handleMountWithTTLFlag(w, r, client.PrevNoExist); err != nil {
+		httpError(w, "Could not publish mount information", err)
+		return
+	}
+}
+
+func (d daemonConfig) handleMountReport(w http.ResponseWriter, r *http.Request) {
+	if err := d.handleMountWithTTLFlag(w, r, client.PrevExist); err != nil {
 		httpError(w, "Could not publish mount information", err)
 		return
 	}
--- a/volplugin/handlers.go
+++ b/volplugin/handlers.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
-	"os"
+	"sync"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/volplugin/config"
@@ -18,6 +18,11 @@ type unmarshalledConfig struct {
 	Name    string
 	Tenant  string
 }
+
+var (
+	mountStopChans = map[string]chan struct{}{}
+	mountMutex     = sync.Mutex{}
+)
 
 func nilAction(w http.ResponseWriter, r *http.Request) {
 	content, err := json.Marshal(VolumeResponse{})
@@ -140,7 +145,7 @@ func getPath(master string) func(http.ResponseWriter, *http.Request) {
 	}
 }
 
-func mount(master, host string) func(http.ResponseWriter, *http.Request) {
+func mount(master, host string, ttl int) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		uc, err := unmarshalAndCheck(w, r)
 		if err != nil {
@@ -167,6 +172,9 @@ func mount(master, host string) func(http.ResponseWriter, *http.Request) {
 			httpError(w, "Reporting mount to master", err)
 			return
 		}
+
+		stopChan := addStopChan(uc.Request.Name)
+		go heartbeatMount(master, ttl, ut, stopChan)
 
 		driverOpts := storage.DriverOptions{
 			Volume: storage.Volume{
@@ -196,7 +204,7 @@ func mount(master, host string) func(http.ResponseWriter, *http.Request) {
 	}
 }
 
-func unmount(master string) func(http.ResponseWriter, *http.Request) {
+func unmount(master, host string) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		uc, err := unmarshalAndCheck(w, r)
 		if err != nil {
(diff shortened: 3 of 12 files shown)


volmaster/daemon.go
```diff
@@ -129,16 +134,32 @@ func (d daemonConfig) handleUnmount(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		httpError(w, "Unmarshalling request", err)
-		return
+		return err
 	}
 
 	req.Reason = "Mount"
 
-	if err := d.config.PublishUse(req); err != nil {
+	fmt.Println(req)
```
mapuri: why not use `log` facility?

volmaster/daemon.go
```diff
@@ -129,16 +134,32 @@ func (d daemonConfig) handleUnmount(w http.ResponseWriter, r *http.Request) {
...
+
+	if err := d.config.PublishUseWithTTL(req, time.Duration(d.mountTTL)*time.Second, exist); err != nil {
```
mapuri: This is just a thought that I got while reviewing this PR. I thought to bring it up in case you already didn't consider and rejected it due to other implications that I might have missed.

I think using the `d.mountTTL` (passed to volmaster as a flag) creates a dependency between ttl values set for volplugin and volmaster at their startup, for instance if I got it right we need to set volplugin ttl to be smaller (or equal to) that volmaster's ttl for the current approach to mostly work. I think part of this comes from that fact that we let volmaster update the runtime state for a mount created by volplugin in etcd.

Alternatively, would it be simpler if we let volplugin update this state in etcd itself? So volmaster stores the configuration state of volume and distributes it to volplugin. While volplugin's update the runtime mount state in etcd. In an event where two volplugin instances try to create same state I think we can let etcd fail one of the writes.

volplugin/handlers.go
```diff
@@ -168,6 +173,9 @@ func mount(master, host string) func(http.ResponseWriter, *http.Request) {
 			return
 		}
 
+		stopChan := addStopChan(uc.Request.Name)
+		go heartbeatMount(master, ttl, ut, stopChan)
```
mapuri: We should report a mount  and start the ttl thread only after `driver.Mount` call below succeeds, right?

volplugin/utils.go
```diff
@@ -164,3 +189,26 @@ func splitPath(name string) (string, string, error) {
 func joinPath(tenant, name string) string {
 	return strings.Join([]string{tenant, name}, ".")
 }
+
+func addStopChan(name string) chan struct{} {
+	mountMutex.Lock()
+	stopChan := make(chan struct{}, 1)
+
+	if sc, ok := mountStopChans[name]; ok {
```
mapuri: shouldn't it be an error if a channel already exists with that name? I might being a bit conservative here :)

volplugin/utils.go
```diff
@@ -164,3 +189,26 @@ func splitPath(name string) (string, string, error) {
...
+		sc <- struct{}{}
+	}
+
+	mountStopChans[name] = stopChan
+	mountMutex.Unlock()
```
mapuri: we can perhaps defer the call to unlock here?

volplugin/utils.go
```diff
@@ -164,3 +189,26 @@ func splitPath(name string) (string, string, error) {
...
+
+	return stopChan
+}
+
+func removeStopChan(name string) {
+	if sc, ok := mountStopChans[name]; ok {
+		sc <- struct{}{}
+		mountMutex.Lock()
```
mapuri: Unless I am reading this wrong, I am wondering if we shall take the `Lock` before the if statement itself, otherwise there seems a race where two go routines can still try to delete the same entry and step over each other?

volplugin/volplugin.go
```diff
@@ -95,3 +110,45 @@ func logHandler(name string, debug bool, actionFunc func(http.ResponseWriter, *h
...
+		parts := strings.Split(mount.Volume.Name, "/")
+		if len(parts) != 2 {
+			log.Warnf("Invalid volume named %q in ceph scan: skipping refresh", mount.Volume.Name)
+			continue
+		}
+
+		log.Infof("Refreshing existing mount for %q", mount.Volume.Name)
+
+		volConfig, err := requestVolumeConfig(dc.Master, parts[0], parts[1])
+		if err != nil {
```
mapuri: I am trying to understand the conditions when the err will be not nil here, in order to verify the possible handling. Let me know if I got it right.

One of the conditions when this can happen is when a volume was deleted (from volmaster) while volplugin was down. In that case we might just want to cleanup the mount and move on.

Another would be when volmaster itself is unreachable. In that case we might not be able to do much but just return error and wait for instructions from volmaster when it comes up like it can reply whatever it has.

Is there any other error condition that we need to think about?

Makefile
```diff
@@ -57,7 +57,7 @@ unit-test-nocoverage:
 unit-test-nocoverage-host: golint-host govet-host
 	HOST_TEST=1 GOGC=1000 godep go test -v ./... -check.v
 
-build: golint
+build: golint govet
```
unclejack: This seems to have sneaked in. Would you be OK with having this in a separate PR which makes more changes to the makefile? I don't have a strong opinion about this, but we should avoid this when possible.

erikh: It got balled up because of a squash, but it needs to be in there. Can 
we take a pass on this? I think this is a small enough change to not 
worry about it too much.

volmaster/daemon.go
erikh: whoops; snuck in there.

volmaster/daemon.go
erikh: I agree, but volplugin handling the updates is less desirable; this allows volmasters to work directly against etcd without sharing credentials that the volmaster/volplugin may share, ensuring a security model with cert auth and so forth.

I would argue we leave this for now, and introduce some kind of general configuration for what will eventually be called "Contiv Storage". This configuration could live in etcd and be passed through the REST interface so that volcli and volplugin can interact with the volmaster to get and change this information.

I can put in an issue for this now if it seems to you like a reasonable replacement. The TTL (and other settings, like debug) could be stored there so it could be referenced by both services.

volplugin/handlers.go
erikh: So, this is an interesting first-responder problem. Basically, if we don't acquire the lock first, someone else can try to mount it and succeed. If we acquire the lock first, the next attempt to mount will break allowing the functionality we want.

Basically this defeats a mount race across the cluster. I don't believe there's a better way to do it, but I'm open to suggestions.

volplugin/utils.go
erikh: Ah, this is some confusing logic I should clean up or comment.

Basically, if we already have a stop channel there we want to send a message JUST IN CASE to clean up the goroutine before we replace it with the new channel and goroutine.

volplugin/utils.go
erikh: ok, will do.

volplugin/utils.go
erikh: You're right. Will fix.

volplugin/volplugin.go
erikh: Hmm. I actually totally overlooked the volmaster down state. I will look into dividing these errors up, it will greatly clarify what needs to be done as you have suggested.

Another condition is when the volume didn't exist in the first place, but is registered as a legal mount point we detected at startup. This is much harder to deal with as there's no accessible journal of when it changed. I'm not entirely sure if we should be handling those at all or taking action to reconcile the system with what we perceive is reality by removing the mounts.

volmaster/daemon.go
mapuri: sounds good, yes having a centralized config for ttl solves the issue too.

volplugin/handlers.go
mapuri: interesting. I didn't think about this, that's a bit harder problem. 

At initial thought I wondered if we could depend on ceph to fail a duplicate mount call. But I think that won't work if the both the competing mount calls are made from same host?

We can leave this as is until we find a good solution, but might be good to add a comment just to remind us in future :)

volplugin/utils.go
mapuri: yeah, I got that but I was wondering if it's even ok/expected to have an existing stopChannel and getting a another request to add it.

volplugin/handlers.go
erikh: We could use rbd locks to accomplish this. I can put in a ticket, but it's really out of scope (and a large change).

volplugin/utils.go
erikh: Ideally, no. However, it's kind of protecting itself against bugs which might leak goroutines this way, so I thought it was appropriate.

volplugin/handlers.go
erikh: Oh, FTR, ceph will happily map rbd devices anywhere for N hosts, so we keep these locks for a very useful reason. :)

mapuri: ok, sgtm... I am fine leaving this as is or filing a ticket to explore using rbd locks.

volplugin/utils.go
mapuri: sure, thanks for clarifying. This works, I was just being conservative about missing out catching any leaked routines.

volplugin/volplugin.go
mapuri: > I'm not entirely sure if we should be handling those at all or taking action to reconcile the system with what we perceive is reality by removing the mounts.

I would not try to handle this if it's not going to be a common case. If we can assume that ceph is managed by volplugin it will simplify error handling.

erikh:
PTAL, I amended the commit to take in some of your comments and fix some testing issues as well.

mapuri:
thanks @erikh, LGTM

erikh:
Going to merge this @unclejack, since you will be out.

shaleman:
Couple of comments:

 1. When volplugin restarts, it is possible that etcd has TTL timed out previous `use` entry. You might need a way to handle it. I'm not sure relying on PrevExist is a good idea.
 2. When network interface on volplugin host is down for longer than TTL period, someone else could acquire the locks and mount the same volume. If network connectivity comes back up, two hosts might end up mounting the same volume. So, volplugin should unmount the volumes if it can't refresh TTL.
 3. More generic version of above is, its possible to loose a lock even if volplugin thinks its refreshing it. 
 4. If volume config could change over time(E.g. snapshot frequency), i'm not sure if comparing previous config while refreshing TTL is good idea. just using volume's unique identifier might be better.
 5. As @mapuri mentioned, Having to go thru volmaster for acquiring distributed locks has the downside that you have to handle volmaster node failure. 

-sukhesh

erikh:
oh crap! I just merged it seconds ago.

A couple of things in the works to deal with this:

I think rbd locks will solve #2. By acquiring the lock the mons should still know that it is in use, even if the host disappears for a few minutes. At least, to put it another way, ceph will have problems before volplugin will, with the rbd locks in place. At that point a restart of volplugin will be the only thing needed to re-register the mount, which should not be acquired elsewhere.

Failure of ceph is a problem we need to address, however.

Right now volume config can't drift. We can address this when that is necessary, probably by partitioning that option data out.

Handling volmaster node failure is ok; it runs multi-master so we can write some failover code there eventually.

Thanks for this! Gave me a lot to think about (and issues to put in)

PR #82, opened 2015-12-10
More tests

erikh:
These tests push the tests for `config` into the 80+ percentile for coverage. The outliers are marshal calls (hard to test and somewhat pointless) and some output validation routines that were particularly hard to trigger due to how things are done in volplugin.

Note that the tests will not pass right now; when @mapuri's ansible etcd changes can be integrated it will pass.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"path"
 
+	"github.com/alecthomas/units"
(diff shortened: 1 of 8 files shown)


erikh:
ok guys, this should pass and is ready for review, if you have time please do @mapuri @unclejack

PR #83, opened 2015-12-10
Upgrade ansible

erikh:
This brings in the latest ansible from https://github.com/contiv/ansible. Merging after tests pass; no need to review.


Makefile
```diff
@@ -130,4 +130,4 @@ release: tar
 	@make clean-tar
 
 vendor-ansible:
-	git subtree pull --prefix ansible https://github.com/contiv/ansible HEAD --squash
```
mapuri: @erikh, you might not be expecting a comment on this but this just caught my eye and I wasn't sure if this is intentional so thought will check :).

I think  `--squash` flag might be useful to have to avoid the main projects history from being polluted by the vendored projects history?

erikh: `--squash` causes a lot of problems. it can't trace back the commits that generated the squash, so the next pull breaks.

At least, this is what I am seeing in volplugin. Note https://github.com/erikh/volplugin/commit/8b2674fdb7bd8bb9dda3c2417d1bb32fbbe13bc8 which had to be reverted because of this.

PR #86, opened 2015-12-12
More systemtests

erikh:
This adds two new battery tests that exercise parallel operations on the system. 

They're don't work very hard because vagrantssh still doesn't scale beyond ~20 connections, need to figure out what's wrong with the sshd config which has to be the issue at this point.

/cc @mapuri @unclejack

### DIFF
--- a/systemtests/util_test.go
+++ b/systemtests/util_test.go
@@ -62,7 +62,7 @@ func (s *systemtestSuite) createVolumeHost(tenant, host string, opts map[string]
 }
 
 func (s *systemtestSuite) createVolume(host, tenant, name string, opts map[string]string) error {
-	log.Infof("Creating %s/%s on %s", tenant, name, host)
+	log.Infof("Creating %s/%s on %q", tenant, name, host)
 
 	optsStr := []string{}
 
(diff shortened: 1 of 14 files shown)


systemtests/util_test.go
```diff
@@ -138,14 +138,14 @@ func (s *systemtestSuite) uploadIntent(tenantName, fileName string) (string, err
 	return s.volcli(fmt.Sprintf("tenant upload %s < /testdata/%s.json", tenantName, fileName))
 }
 
-func (s *systemtestSuite) pullUbuntu() error {
+func (s *systemtestSuite) pullDebian() error {
 	wg := sync.WaitGroup{}
 	errChan := make(chan error, 3)
-	for _, host := range []string{"mon0", "mon1", "mon2"} {
+	for _, host := range s.vagrant.GetNodes() {
```
mapuri: we can possibly use `SSHExecAllNodes()` function here.

mapuri:
LGTM, otherthan one minor comment.

systemtests/util_test.go
erikh: Hmm... yeah, that should work.

PR #93, opened 2015-12-21
systemtests: fix behavior of CONTIV_SOE to better match netplugin's concept

erikh:


systemtests/init_test.go
```diff
@@ -31,15 +31,16 @@ func (s *systemtestSuite) SetUpTest(c *C) {
 }
 
 func (s *systemtestSuite) TearDownTest(c *C) {
-	if os.Getenv("CONTIV_SOE") != "" {
-		log.Infof("SOE set. Terminating immediately")
+	if c.Failed() && os.Getenv("CONTIV_SOE") != "" {
```
mapuri: @erikh I have a similar requirement for cluster-mgr systemt-tests. And I faced an issue using `c.Failed()` method in `TearDownTest()`. It appears that the test status is not passed to the TearDownTest so `c.Failed()` always returns false here.

Did you happen to hit this? For now the way I have worked-around this problem is by adding a Assert() function (https://github.com/mapuri/cluster/blob/systemtests/management/src/systemtests/cli_test.go#L30) of my own to wrap the `c.Assert()` and use that in TearDownTest() (https://github.com/mapuri/cluster/blob/systemtests/management/src/systemtests/cli_test.go#L72). But I am hoping there is a better solution.

erikh: Huh. Maybe I forgot to test this branch as well. Perhaps I should wait 
on this one.

PR #96, opened 2015-12-24
Use the rbd showmapped command after mapping, as rbd outputs nothing.

slepp:
On the distribution of Ceph I'm using, rbd outputs nothing when it maps. Using the showmapped command after the map returns gives the JSON result, which is iterated for the new device name.

### DIFF
--- a/storage/backend/ceph/internals.go
+++ b/storage/backend/ceph/internals.go
@@ -6,14 +6,30 @@ import (
 	"regexp"
 	"strings"
 
+	"encoding/json"
+
 	"github.com/contiv/volplugin/storage"
 
 	log "github.com/Sirupsen/logrus"
 )
 
 func (c *Driver) mapImage(do storage.DriverOptions) (string, error) {
-	blkdev, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
-	device := strings.TrimSpace(string(blkdev))
+	var outputdata map[string]interface{}
+	var device string
+
+	_, err := exec.Command("rbd", "map", do.Volume.Name, "--id", "contiv", "--pool", do.Volume.Params["pool"]).Output()
+	output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
+
+	json.Unmarshal(output, &outputdata)
+	fmt.Println(outputdata)
+
+	device = ""
+
+	for i := range outputdata {
+		if outputdata[i].(map[string]interface{})["name"].(string) == do.Volume.Name {
+			device = outputdata[i].(map[string]interface{})["device"].(string)
+		}
+	}
 
 	if err == nil {
 		log.Debugf("mapped volume %q as %q", do.Volume.Name, device)
--- a/storage/backend/ceph/internals.go
+++ b/storage/backend/ceph/internals.go
@@ -20,38 +20,41 @@ func (c *Driver) mapImage(do storage.DriverOptions) (string, error) {
 		Device string `json:"device"`
 	}
 
-	var device string
-
 	blkdev, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
-	device = strings.TrimSpace(string(blkdev))
+	device := strings.TrimSpace(string(blkdev))
 
-	if device == "" && err == nil {
-		output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
-
-		if err == nil {
-			err := json.Unmarshal(output, &rbdMaps)
-			if err == nil {
-				for i := range rbdMaps {
-					if rbdMaps[i].Name == do.Volume.Name && rbdMaps[i].Pool == do.Volume.Params["pool"] {
-						device = rbdMaps[i].Device
-						break
-					}
-				}
-				if device == "" {
-					return "", fmt.Errorf("Volume %s in pool %s not found in RBD showmapped output", do.Volume.Name, do.Volume.Params["pool"])
-				}
-			} else {
-				return "", fmt.Errorf("Could not parse RBD showmapped output: %s", output)
-			}
-		}
+	if err != nil {
 		return "", err
 	}
 
-	if err == nil {
-		log.Debugf("mapped volume %q as %q", do.Volume.Name, device)
+	if device == "" {
+		output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
+
+		if err != nil {
+			return "", err
+		}
+
+		err = json.Unmarshal(output, &rbdMaps)
+
+		if err != nil {
+			return "", fmt.Errorf("Could not parse RBD showmapped output: %s", output)
+		}
+
+		for i := range rbdMaps {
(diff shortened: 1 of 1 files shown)


storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+
 	"github.com/contiv/volplugin/storage"
 
 	log "github.com/Sirupsen/logrus"
 )
 
 func (c *Driver) mapImage(do storage.DriverOptions) (string, error) {
-	blkdev, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
-	device := strings.TrimSpace(string(blkdev))
+	var outputdata map[string]interface{}
```
mapuri: Instead of a `map[string]interface{}`, I think it might be better to define `outputdata` to be a struct, this will avoid the dictionary lookup and type assertion below at line 29 and hence a chance of panic.

I am not sure what the json output looks like, but assuming it is a array of volume infos, we might be able to do something like this:

```
var outputData struct {
    volumes []struct {
        Name string `json:"name"`
    } `json:"volumes"`
}

json.Unmarshal(output, &outputData)
...
...

if outputdata[i].Name != do.Volume.Name {
...
}
```

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+	var device string
+
+	_, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
```
mapuri: I think we should check for error here and not proceed if the command failed.

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+	output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
+
+	json.Unmarshal(output, &outputdata)
```
mapuri: we should check for unmarshal errors here

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+	fmt.Println(outputdata)
```
mapuri: we should log this using `log.Debugf()` function instead

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+
+	device = ""
+
+	for i := range outputdata {
+		if outputdata[i].(map[string]interface{})["name"].(string) == do.Volume.Name {
```
mapuri: if we are not able to define a struct for outputdata as suggested above we should check for key lookups and type assertion errors to avoid a panic.

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+			device = outputdata[i].(map[string]interface{})["device"].(string)
+		}
```
mapuri: we can `break` out of the loop once we have found the matching volume.

mapuri:
@slepp thanks for the patch. I am not much familiar with ceph command line so will let @erikh comment on the approach as well. Other than that I have a few minor comments.

slepp:
Thanks for all the feedback. It's been about three years since I last wrote anything in Go, so it's a bit rusty. I'll review and make those changes in the next day or two. The struct makes sense and seems to be very constant and simple in format, so should do well in this context. I'll also add a check if rbd outputs the volume name as the original code used, for compatibility.

erikh:
@slepp for my curiosity, which version of ceph are you running?

erikh:
@slepp If you need help with Go, email me at the address on my github profile and we can work out a way to discuss this in a faster timeframe.

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,30 @@ import (
...
+	var outputdata map[string]interface{}
+	var device string
+
+	_, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
+	output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
+
+	json.Unmarshal(output, &outputdata)
+	fmt.Println(outputdata)
+
+	device = ""
```
erikh: This line is not necessary because it will be the default value of the declared `device` above.

slepp:
@erikh I'm using Ceph version 0.80.11 on the server running rbd.

An example JSON output is:

{"0":{"pool":"rbd","name":"test1","snap":"-","device":"\/dev\/rbd0"}}

I've updated things now with the suggestions you and @mapuri made.

erikh:
Ok, will review later today. Thank you so much!

erikh:
I'm sorry to ask this; if you could do this with your branch to kick the CI service:

```
git commit -a -s --amend
git push -f <my remote> <my branch>
```

The ansible used to set up the boxes occasionally fails due to some proxy issues where our Jenkins service lives. We're working on it. :)

slepp:
@erikh CI service kicked (hopefully :).

erikh:
Ok, sorry to do this after all the CI stuff as I didn't have time to look it over earlier.

Please test each error as you are, but when you get an error please return it, like so:

```go
return "", err
```

Otherwise this patch LGTM. @mapuri care to comment?

slepp:
Updated with error returns, including if the mapping isn't found (it would've cascaded out silently before).

Not sure if CI needs a kick or not.

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,46 @@ import (
...
+	var device string
+
 	blkdev, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
-	device := strings.TrimSpace(string(blkdev))
+	device = strings.TrimSpace(string(blkdev))
+
+	if device == "" && err == nil {
+		output, err := exec.Command("rbd", "showmapped", "--format", "json").Output()
+
+		if err == nil {
```
mapuri: this is just a minor code styling nit, by doing a `err != nil` check earlier, the large part of the code within the `if` can be kept indented to left, which is also easier t read.

```
if err != nil {
    return "", err
}

... the main logic left indented goes here
...
```

storage/backend/ceph/internals.go
```diff
@@ -6,14 +6,46 @@ import (
...
+			err := json.Unmarshal(output, &rbdMaps)
+			if err == nil {
```
mapuri: similar to comment above, if you check for `err != nil` and return the else block can be removed and the looping code will be indented to left making it more readable and less nested.

mapuri:
@slepp thanks for taking care of my comments. The code changes look good to me. I have a few minor comments around code styling, please see if you can incorporate those. Thanks again for your time to address comments!

slepp:
Updated the code there for readability (it was a bit confusing before). Moved the err check for the initial map command back near the top where it should actually be checked. Final return changed to always be device, nil, as we can't end up here with err != nil.

mapuri:
thanks @slepp! changes LGTM

erikh:
LGTM

erikh:
Thanks for your contribution!

PR #97, opened 2015-12-28
minor corrections in readme

vijayvikrant:
`make run` will run both the volplugin and volmaster on all the hosts. so, the statement about "starting volplugin" manually is not necessary.
make start will fail if https_proxy is specified. so, adding a note for not using https_proxy as ansible does not support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,10 @@ Your guests will configure themselves.
 
 Be sure to start and run the environment with `make start` before you
 continue with these steps. You must have working vagrant, virtualbox, and
-ansible.
+ansible. If you are behind a proxy server, then use only http proxy.
 
 These instructions ssh you into the `mon0` vm. If you wish to test the
-cross-host functionality, ssh into `mon1` or `mon2` with `vagrant ssh`. Then
-start at the "starting volplugin" (not volmaster) step.
+cross-host functionality, ssh into `mon1` or `mon2` with `vagrant ssh`.
 
 1. Run the suite: `make run`.
 1. SSH into the host: `make ssh`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,11 +55,10 @@ Your guests will configure themselves.
 
 Be sure to start and run the environment with `make start` before you
 continue with these steps. You must have working vagrant, virtualbox, and
-ansible.
+ansible. If you are behind a proxy server, then use only http proxy.
```
erikh: I'm not sure what this means, can you clarify it? Maybe reference `http_proxy` and `https_proxy` as ways to configure the makefile.

erikh:
the rest LGTM. please correct the issue.

README.md
vijayvikrant: @erikh: Ansible task will fail with https proxy.
TASK: [ceph-common | install the ceph stable repository key] ******************
failed: [mon1] => {"failed": true}
msg: Unsupported proxy scheme: https. Currently ansible only supports HTTP proxies.
failed: [mon0] => {"failed": true}
msg: Unsupported proxy scheme: https. Currently ansible only supports HTTP proxies.
failed: [mon2] => {"failed": true}
msg: Unsupported proxy scheme: https. Currently ansible only supports HTTP proxies.

FATAL: all hosts have already failed -- aborting

vijayvikrant: When i see this i do `export https_proxy=`echo $http_proxy``

vijayvikrant:
ping @erikh 
added the ansible defect link to the readme and made the proxy related instructions more explicit.

PR #99, opened 2016-01-03
Vmware Support

erikh:
@mapuri @unclejack PTAL

Note that this requires the VMware fusion or workstation which is a for-pay addition by hashicorp.

The tests run significantly faster, and provisioning time is improved if you do use it. The networking is also marginally more sane. :)

Other patches have been submitted to the respective repositories so they can be added that way.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -25,7 +25,7 @@
 		},
 		{
 			"ImportPath": "github.com/contiv/vagrantssh",
-			"Rev": "2b44706f52b6a04d0441e6f0938e13a3e1f0bd49"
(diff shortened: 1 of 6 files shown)


mapuri:
@erikh 

LGTM

Given vmware support is a paid thing so I think we will need to keep supporting virtualbox based environment for our users. Is that right?

erikh:
On Sun, Jan 03, 2016 at 10:45:08AM -0800, Madhav Puri wrote:
> Given vmware support is a paid thing so I think we will need to keep supporting virtualbox based environment for our users. Is that right?

Very correct.

PR #101, opened 2016-01-03
Rbd locks

### LINKED ISSUE #79: Implement RBD locking
Questions arising from #78 seemed to indicate that implementing mandatory rbd locking would solve a few problems acquiring locks in circumstances where network partitions were present.

Implementing this directly in the storage layer is the way to go. It should not be part of the driver interface, but instead be an internal component to the ceph driver.

erikh:
This implements RBD locking in the ceph backend.

closes #79

### DIFF
--- a/storage/backend/ceph/internals.go
+++ b/storage/backend/ceph/internals.go
@@ -1,26 +1,69 @@
 package ceph
 
 import (
+	"encoding/json"
 	"fmt"
 	"os/exec"
 	"regexp"
 	"strings"
-
-	"encoding/json"
 
 	"github.com/contiv/volplugin/storage"
 
 	log "github.com/Sirupsen/logrus"
 )
 
+func (c *Driver) lockImage(do storage.DriverOptions) error {
+	poolName := do.Volume.Params["pool"]
+
+	if err := exec.Command("rbd", "lock", "add", do.Volume.Name, do.Volume.Name, "--pool", poolName).Run(); err != nil {
+		log.Debugf("Could not acquire lock for %q: %v", do.Volume.Name, err)
+		return err
+	}
+
+	return nil
+}
+
+func (c *Driver) unlockImage(do storage.DriverOptions) error {
+	poolName := do.Volume.Params["pool"]
+
+	output, err := exec.Command("rbd", "lock", "--format", "json", "list", do.Volume.Name, "--pool", poolName).Output()
+	if err != nil {
+		return fmt.Errorf("Error running `rbd lock list` for volume %q: %v", do.Volume.Name, err)
+	}
+
+	locks := map[string]map[string]string{}
+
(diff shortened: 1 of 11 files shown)


erikh:
@mapuri @unclejack can you review this?

storage/backend/ceph/internals.go
```diff
@@ -1,26 +1,32 @@
...
+	var rbdMaps map[string]struct {
+		Pool   string `json:"pool"`
+		Name   string `json:"name"`
 		Device string `json:"device"`
 	}
 
-	blkdev, err := exec.Command("rbd", "map", do.Volume.Name, "--pool", do.Volume.Params["pool"]).Output()
+	poolName := do.Volume.Params["pool"]
+
+	if err := exec.Command("rbd", "lock", "add", do.Volume.Name, do.Volume.Name, "--pool", poolName).Run(); err != nil {
```
mapuri: just for my understanding, does the call to `lock` associates the lock for the pool to the machine/host from which this operation is performed? I am wondering if unlock needs to be called from the same host or anyone calling unlock will unlock the pool.

storage/backend/ceph/internals.go
```diff
@@ -1,26 +1,32 @@
...
+		log.Debugf("Could not acquire lock for %q: %v", do.Volume.Name, err)
+		return "", err
+	}
+
```
mapuri: just wondering if we shall defer a unlock operation in case one of the operations later fail?

erikh: hmm... probably a good idea!

storage/backend/ceph/internals.go
erikh: It's been a few weeks since I wrote this, but I am pretty sure that the unmount driver API call will cause the unlock to happen on the same host; but I will verify before submitting updates for the rest.

erikh:
@unclejack @mapuri PTAL

mapuri:
LGTM

I am wondering if it we need to add any tests here. Are the following scenarios valid and should be tested?

- add a test to verify that a lock is indeed taken when a volume is created by volplugin.
- some negative cases like trying volume delete/create using ceph commands once it is created by volplugin. This shall fail I assume.

erikh:
I think the first tests is indeed necessary (in the unit tests) and will correct.

The second one is already tested I think, but I will verify tonight.

mapuri:
thanks @erikh, sgtm. It still looks good from my end.

erikh:
I can verify the second test is already tested, but it triggers on the use lock, not the rbd lock. I think we can consider it tested in the systemtests this way, but it's not perfect. The unfortunate problem is that the use locks are checked before the rbd locks can interfere, since they are an back-up safeguard, not something we explicitly rely on.

I am working on the other unit test now.

PR #106, opened 2016-01-20
Normalize volcli volume commands

### LINKED ISSUE #65: volcli: use `tenant/volume` syntax instead of multiple args
`volcli` has many tools that use this notation: `volcli command subcommand tenant volume`, which I think would be more canonically represented as `volcli command subcommand tenant/volume` like it is used inside docker.

erikh:
fixes #65


erikh:
PTAL @mapuri @unclejack

volcli/volcli.go
```diff
@@ -125,7 +135,7 @@ func TenantList(ctx *cli.Context) {
 // VolumeCreate creates a new volume with a JSON specification to store its
 // information.
 func VolumeCreate(ctx *cli.Context) {
-	if len(ctx.Args()) != 2 {
+	if len(ctx.Args()) != 1 {
 		errExit(ctx, fmt.Errorf("Invalid arguments"), true)
```
mapuri: I understand that this comment is unrelated to the PR but thought will share since I noticed it :) but feel free to track it through separate issue/PR. 

This error might be more useful if it states something like `volcli expects only one argument` ? Same is true at multiple places below. May be we could use a variable/constant here.

volcli/volcli.go
```diff
@@ -315,7 +335,7 @@ func UseGet(ctx *cli.Context) {
 // UseTheForce deletes the use entry from etcd; useful for clearing a
 // stale mount.
 func UseTheForce(ctx *cli.Context) {
-	if len(ctx.Args()) != 2 {
+	if len(ctx.Args()) != 1 {
```
mapuri: I think the `ArgUsage` string update for `get` and `force-remove` commands is missing above. Sorry, I could post comment there as github only let's me comment on changed lines.

mapuri:
LGTM other than few minor comments

volcli/volcli.go
erikh: Yeah, I think this is an artifact of me doing this too quickly initially. #108 was just created so this can be corrected.

erikh:
I rebased this on top of #110 and #107, so we need to merge those first. However, I think all the concerns here have been resolved in f9ce138.

volcli/volcli.go
mapuri: +1, tracking this separately with #108 sgtm

mapuri:
LGTM

PR #107, opened 2016-01-20
Makefile: run after build

### LINKED ISSUE #98: `make start` should also `make run`
/cc @jainvipin

erikh:
fixes #98

PTAL @jainvipin @mapuri @unclejack

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ unit-test-nocoverage-host: golint-host govet-host
 
 build: golint govet
 	vagrant ssh mon0 -c 'sudo -i sh -c "cd $(GUESTGOPATH); make run-build"'
+	make run
 
 run:
 	@set -e; for i in $$(seq 0 2); do vagrant ssh mon$$i -c 'cd $(GUESTGOPATH) && make run-volplugin run-volmaster'; done


Makefile
```diff
@@ -64,6 +64,7 @@ unit-test-nocoverage-host: golint-host govet-host
 
 build: golint govet
 	vagrant ssh mon0 -c 'sudo -i sh -c "cd $(GUESTGOPATH); make run-build"'
+	make run
```
mapuri: @erikh from #98 looks like @jainvipin wanted `make run` to be added to `start` target but I see you have added it to `build` target instead. Is this intended?

erikh: Yes; I thought it made more sense there, since after a build you will want to restart it anyways, for manual testing purposes.

mapuri: oh, k. I think he might have been looking for `make demo` kind of target of netplugin which not only create a vagrant setup but also starts the netplugin/netmaster etc

But I will let him comment, otherwise sgtm.

On a unrelated note, bringing up volplugin might interfere with system-tests but I think sanity will tell us :)

erikh: Yeah, `make build` is run by both the systemtests and `make start`, so the `run` bits will get applied at the right times.

The systemtests already accounted for this, but I validated that first.

PR #110, opened 2016-01-21
Fix a few things

erikh:
Sorry for the vague description, but this fixes a number of build-related issues with volplugin that ... aren't related to volplugin itself.

The commit messages speak for themselves.

/cc @mapuri @unclejack

### DIFF
--- a/ansible/roles/base/tasks/redhat_tasks.yml
+++ b/ansible/roles/base/tasks/redhat_tasks.yml
@@ -30,14 +30,15 @@
 - name: install and start ntp
   shell: systemctl enable ntpd
 
-- name: install python-crypto
-  yum: name=python-crypto state=present
-  register: python_crypto_result
+- name: install ansible
+  yum: name=ansible state=present
+  register: ansible_result
   ignore_errors: yes
 
 - name: remove python crypt egg file to work-around https://bugs.centos.org/view.php?id=9896&nbn=2
-  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-2.6.1-py2.7.egg-info
-  when: python_crypto_result.msg == "Error unpacking rpm package python-crypto-2.6.1-1.el7.centos.x86_64\n"
+  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-*.egg-info
+  when: '"Error unpacking rpm package python2-crypto-" in ansible_result.msg'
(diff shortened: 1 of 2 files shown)


erikh:
all open PRs are bringing this in now, so please review this first, when you have the time of course.

ansible/roles/base/tasks/redhat_tasks.yml
```diff
@@ -30,14 +30,15 @@
 - name: install and start ntp
   shell: systemctl enable ntpd
 
-- name: install python-crypto
-  yum: name=python-crypto state=present
```
mapuri: Since the issue is with python-crypto installation I had added this task, which kind of allowed me have just one task for ansible installation with no conditionals and follow error-checking.

Note that if ansible installation were to fail for a reason other than python-crypto, the current change will cause the palybook to move on and not fail which is something we do not want.

ansible/roles/base/tasks/redhat_tasks.yml
```diff
@@ -30,14 +30,15 @@
...
-  register: python_crypto_result
+- name: install ansible
+  yum: name=ansible state=present
+  register: ansible_result
   ignore_errors: yes
 
 - name: remove python crypt egg file to work-around https://bugs.centos.org/view.php?id=9896&nbn=2
-  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-2.6.1-py2.7.egg-info
-  when: python_crypto_result.msg == "Error unpacking rpm package python-crypto-2.6.1-1.el7.centos.x86_64\n"
+  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-*.egg-info
```
mapuri: @erikh, 
the change looks fine. But I am just curious about making the string match more generic, did you hit this issue with a different version of python-crypto package? Since this was a workaround I had intentionally kept it for a very specific version.

erikh: yes, on centos72 the package is `python2-crypto` because python3. this will affect everything as soon as netplugin upgrades.

erikh:
merged; I hope this is ok @mapuri

mapuri:
sure @erikh , I think the comment on handling `ansible task` failure might need to be addressed but may be we can address once this makes it to ansible repo.

PR #112, opened 2016-01-22
systemtests: wait for processes to be up

unclejack:


mapuri:
@unclejack I think you can use one the existing utility functions in systemtest-utils package `ServiceStartAndWaitForUp()`

unclejack:
@mapuri We're not using services for these and we can't use that.

mapuri:
I see, I think we should be using services for testing but that is a different issue I guess :)

There is a more generic `WaitForDone()` in systemtest-utils as well that we could possibly use. I am hoping to not have to write new utility functions where systemtests utils already provides something. Feel free to add missing utilities there.

erikh:
I agree, we should do this. @unclejack I can do this if you don't have time.

erikh:
however, this seems to have fixed the issues in the system tests.

unclejack:
@erikh @mapuri: I'll take a look and make the changes.

erikh:
carrying this, I'll preserve your commits @unclejack

PR #115, opened 2016-01-25
Amalgamated PRs

erikh:
This combines #107 #106 and #101 as well as some fixes to make them all work together.

I will close the other PRs now. Please review them commit-by-commit, I attempted to clarify and squash as I could to cleanly separate the different features.


storage/backend/ceph/ceph_test.go
```diff
@@ -182,6 +184,26 @@ func (s *cephSuite) TestMounted(c *C) {
 			},
 		},
 	})
+
+	output, err := exec.Command("sudo", "rbd", "lock", "--format", "json", "list", driverOpts.Volume.Name, "--pool", "rbd").Output()
+	c.Assert(err, IsNil)
```
mapuri: I think logging the command output in case of assert failures can be use. We can use `Commentf()` as the last argument to `Assert()` for the same.

storage/backend/ceph/internals.go
```diff
@@ -1,26 +1,69 @@
...
-
 	"github.com/contiv/volplugin/storage"
 
 	log "github.com/Sirupsen/logrus"
 )
 
+func (c *Driver) lockImage(do storage.DriverOptions) error {
+	poolName := do.Volume.Params["pool"]
+
+	if err := exec.Command("rbd", "lock", "add", do.Volume.Name, do.Volume.Name, "--pool", poolName).Run(); err != nil {
```
mapuri: should we use `Output()` or `CombinedOutput()` here instead of `Run()`? It might be useful to log command output in case of failures.

storage/backend/ceph/internals.go
```diff
@@ -1,26 +1,69 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (c *Driver) unlockImage(do storage.DriverOptions) error {
+	poolName := do.Volume.Params["pool"]
+
+	output, err := exec.Command("rbd", "lock", "--format", "json", "list", do.Volume.Name, "--pool", poolName).Output()
```
mapuri: similar to above, using `CombinedOutput()` might be better as that will log stderr as well. And may be we should also log the output in case of failures.

storage/backend/ceph/internals.go
```diff
@@ -1,26 +1,69 @@
...
+	}
+
+	locks := map[string]map[string]string{}
+
+	if err := json.Unmarshal(output, &locks); err != nil {
+		return fmt.Errorf("Error unmarshalling lock report for volume %q: %v", do.Volume.Name, err)
+	}
+
+	if _, ok := locks[do.Volume.Name]; ok {
+		if err := exec.Command("rbd", "lock", "remove", do.Volume.Name, do.Volume.Name, locks[do.Volume.Name]["locker"], "--pool", poolName).Run(); err != nil {
```
mapuri: similar to above, may be we should use `CombinedOutput()` instead of `Run()` and log the output in case of failures.

mapuri:
LGTM other than a few minor comments around logging command output in case of failures.

storage/backend/ceph/ceph.go
```diff
@@ -60,16 +61,26 @@ func (c *Driver) Create(do storage.DriverOptions) error {
 	return exec.Command("rbd", "create", do.Volume.Name, "--size", strconv.FormatUint(do.Volume.Size, 10), "--pool", poolName).Run()
 }
 
+func (c *Driver) unlockAndLog(do storage.DriverOptions) {
+	// this call might fail, so we don't check the error here. It does log it's
```
unclejack: nit: `its failures`

erikh: welp

unclejack:
LGTM once @mapuri's comments are addressed

mapuri:
@erikh 

As I was mentioning on slack, it's ok to handle logging related comments in a separate PR for all tests together.

So this LGTM.

erikh:
FTR, we're moving towards a unified execution path instead of `exec.Cmd` everywhere, so we can deal with this stuff in a more uniform way.

Merging.

PR #116, opened 2016-01-29
Unified exec pipeline

erikh:
This principally affects the ceph driver but it implements (see first commit) a new package called `executor` which is a supervising and interruptible... a wrapper for exec.Cmd structs.

The second commit is the integration into the ceph driver. It does not do anything with hard timeouts yet, but we have a few other steps that need to be taken before we can start leveraging that feature.

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -1,0 +1,238 @@
+// Package executor implements a high level execution context with monitoring and
+// logging features.
+//
+// Please see Executor for more information.
+package executor
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os/exec"
+	"syscall"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// ExecResult is the result of a Wait() operation and contains various fields
+// related to the post-mortem state of the process such as output and exit
+// status.
+type ExecResult struct {
+	Stdout     string
+	Stderr     string
+	ExitStatus uint32
+	Runtime    time.Duration
+
+	executor *Executor
+}
+
+// Executor is the context used to execute a process. The runtime state is kept
+// here. Please see the struct fields for more information.
+//
+// New() is the appropriate way to initialize this type.
+//
+// No attempt is made to manage concurrent requests to this struct after
+// the program has started.
+type Executor struct {
+	// The interval at which we will log that we are still running.
+	LogInterval time.Duration
+
+	// If a true value is passed into this channel, the process will be forcefully
+	// terminated. A false value terminates only the supervising goroutines, and
+	// is used by Wait().
+	TerminateChan chan bool
+
+	// The function used for logging. Expects a format-style string and trailing args.
+	LogFunc func(string, ...interface{})
+
+	// The stdin as passed to the process.
+	Stdin io.Reader
+
+	timeout time.Duration
+
+	command         *exec.Cmd
+	stdout          io.ReadCloser
+	stderr          io.ReadCloser
+	startTime       time.Time
+	terminateLogger chan struct{}
+}
+
+// New creates a new executor from an *exec.Cmd. You may modify the values
+// before calling Start(). See Executor for more information.
+func New(cmd *exec.Cmd) *Executor {
+	return &Executor{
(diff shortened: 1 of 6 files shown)


executor/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/Sirupsen/logrus"
+)
+
+// ExecResult is the result of a Stop() operation and contains various fields
+// related to the post-mortem state of the process such as output and exit
+// status.
+type ExecResult struct {
+	Stdout     string
+	Stderr     string
+	ExitStatus uint32
```
mapuri: should this be of type `exec.ExitError` ?

executor/executor.go
```diff
@@ -0,0 +1,235 @@
...
+// Executor is the context used to execute a process. The runtime state is kept
+// here. Please see the struct fields for more information.
+//
+// New() is the appropriate way to initialize this type.
+//
+// No attempt is made to manage concurrent requests to this struct after
+// the program has started.
+type Executor struct {
+	// The command passed in through New()
+	Command *exec.Cmd
```
mapuri: I think if it is not expected to change once `New()` is called, we should not export this. Also that way we will also force the caller to get things stdout/stderr from `ExecResult` than from `e.Command`

Similar for fields below as well. I think making them exported would make behavior a bit unsafe/anomalous if they are changed after call to `Start()` or may be we should just add comments stating that these should not be changed once command is started.

executor/executor.go
```diff
@@ -0,0 +1,235 @@
...
+	return 0
+}
+
+// Stop waits for the process and return an ExecResult. The program died due to
+// another problem without returning an exit status, a nil result is yielded.B
+//
+// If the ExecResult is returned, error will be nil, regardless of the
+// ExitError returned by (*exec.Cmd).Wait(). If an error is returned it should
+// be handled appropriately.
+func (e *Executor) Stop() (*ExecResult, error) {
```
mapuri: May be we should call this `Wait()` instead? `Stop()` seems to imply that we are going to kill/stop a running thing.

executor/executor.go
```diff
@@ -0,0 +1,235 @@
...
+
+	res := &ExecResult{
+		executor: e,
+		Stdout:   string(stdout),
+		Stderr:   string(stderr),
+	}
+
+	err = e.Command.Wait()
+
+	e.TerminateChan <- false // signal goroutines to terminate gracefully
```
mapuri: may be we should `defer` this in the beginning itself otherwise if we return early we may leave behind some go routines?

mapuri:
@erikh the changes LGTM. It's a pretty nice wrapper package. I have a few minor nits but otherwise good to merge after sanity.

executor/executor.go
erikh: no, that's deliberately a uint32 so it's easier to check. note that it has a String() and Stop() handles the *exec.ExitError appropriately.

executor/executor.go
erikh: Hm. You're right, I should re-think this.

executor/executor.go
erikh: Good idea.

executor/executor.go
erikh: good idea.

erikh:
PTAL @mapuri. Also, please review @unclejack.

executor/executor.go
mapuri: ah, ok..makes sense

executor/executor.go
```diff
@@ -0,0 +1,238 @@
...
+func (e *Executor) TimeRunning() time.Duration {
+	return time.Now().Sub(e.startTime)
+}
+
+func (e *Executor) terminateTimeout() {
+	time.Sleep(e.timeout)
+	e.TerminateChan <- true
+}
+
+func (e *Executor) waitForWait() {
```
mapuri: this  should be `waitforStop()` I think :)

mapuri:
@erikh 

LGTM, thanks for taking care of my comments

unclejack:
LGTM

PR #118, opened 2016-01-29
Dockerfile,.dockerignore,Makefile: docker image support

erikh:
This allows docker images to be built and pushed to the hub.


mapuri:
LGTM

PR #120, opened 2016-02-01
Volmaster fix map race

### LINKED ISSUE #119: if mkfs fails, use lock is acquired and no volume is written
write the volume after rbd create instead of mkfs.

erikh:
fixes #119. See commits.

/cc @mapuri @unclejack


mapuri:
LGTM.

@erikh 
The sanity didn't run due to CI downtime in morning today. Do you want to kick sanity once before merge?

unclejack:
LGTM

Issue #121, opened 2016-02-01
if volcli volume remove fails, use lock is still held


erikh:
probably just needs a defer in the right spot.

PR #124, opened 2016-02-01
Fix use locks

### LINKED ISSUE #122: if volplugin map/mount fails, rbd lock is not freed in some situations

erikh:
resolves #122 #121. Incorporates #120 so please review that first.

right now this is nearly impossible to test, it literally has to have a broken ceph to happen.

/cc @mapuri @unclejack

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -82,15 +82,19 @@ func (c *Driver) unlockAndLog(do storage.DriverOptions) {
 // Format formats a created volume.
 func (c *Driver) Format(do storage.DriverOptions) error {
 	device, err := c.mapImage(do)
-	if err != nil {
+	if err == ErrLockFailed {
 		c.unlockAndLog(do)
+		return err
+	} else if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


storage/backend/ceph/ceph.go
```diff
@@ -82,16 +82,20 @@ func (c *Driver) unlockAndLog(do storage.DriverOptions) {
 // Format formats a created volume.
 func (c *Driver) Format(do storage.DriverOptions) error {
 	device, err := c.mapImage(do)
-	if err != nil {
+	if err == ErrLockFailed {
```
mapuri: looks like we try to unlock as part of error handling so wondering, if this check should be `err != ErrLockFailed`?

I am also wondering why can't we unlock always whenever `err != nil`, that way we can just `defer` the call to unlock when err is not nil?

erikh: let me look at this patch again. I don't think it's very sound and I created it in a rush.

erikh:
closing this; going to overhaul some of the code in here instead.

erikh:
thanks for catching my sloppiness in review; i'm not sure how I thought this was a good idea.

PR #127, opened 2016-02-03
Fixup storage drivers

### LINKED ISSUE #121: if volcli volume remove fails, use lock is still held

### LINKED ISSUE #122: if volplugin map/mount fails, rbd lock is not freed in some situations

erikh:
This incorporates many changes that improve the stability of volmaster create operations, which was failing in a lot of scenarios on slower machines.

PTAL @mapuri @unclejack, this is a big change so please review carefully.

closes #121 
closes #122

### DIFF
--- a/volmaster/daemon.go
+++ b/volmaster/daemon.go
@@ -243,7 +243,7 @@ func (d daemonConfig) handleCreate(w http.ResponseWriter, r *http.Request) {
 		}()
 
 		do, err := createVolume(tenant, volConfig)
-		if err == storage.ErrVolumeExist { // FIXME move to storage
+		if err == storage.ErrVolumeExist {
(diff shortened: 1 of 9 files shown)


unclejack:
LGTM

volmaster/daemon.go
```diff
@@ -96,11 +97,13 @@ func (d daemonConfig) handleRemove(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := removeVolume(vc); err != nil {
+		d.config.RemoveUse(uc, false)
```
mapuri: may be we should just defer call to `d.config.RemoveUse` from this point on?

mapuri:
a lot of cleanup!! LGTM, other than one minor comment.

PR #130, opened 2016-02-05
storage,volmaster,volsupervisor: add timeout

### LINKED ISSUE #111: timeouts for rbd commands/ceph driver
Currently, if the `rbd` tool hangs, so does our suite of tools. incorporate a sideband timeout that appropriately errors out when this happens, so the administrator can fix the ceph installation (Which is probably already hosed)

unclejack:
Fixes #111

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -71,6 +71,15 @@ func New(cmd *exec.Cmd) *Executor {
 	}
 }
 
+// NewWithTimeout creates a new executor from an *exec.Cmd and a timeout.
+// You may modify the values before calling Start(). See Executor for
+// more information.
+func NewWithTimeout(cmd *exec.Cmd, timeout time.Duration) *Executor {
+	exec := New(cmd)
+	exec.SetTimeout(timeout)
+	return exec
+}
+
 func (e *Executor) String() string {
 	return fmt.Sprintf("%v (%v) (pid: %v)", e.command.Args, e.command.Path, e.PID())
 }
--- a/volmaster/daemon.go
+++ b/volmaster/daemon.go
@@ -22,6 +22,7 @@ import (
 type daemonConfig struct {
 	config   *config.TopLevelConfig
 	mountTTL int
+	timeout  time.Duration
 }
 
 // volume is the json response of a volume. Taken from
@@ -44,8 +45,11 @@ type volumeGet struct {
 }
 
 // Daemon initializes the daemon for use.
-func Daemon(config *config.TopLevelConfig, ttl int, debug bool, listen string) {
-	d := daemonConfig{config, ttl}
+func Daemon(config *config.TopLevelConfig, ttl, timeout int, debug bool, listen string) {
+	d := daemonConfig{config: config,
+		mountTTL: ttl,
+		timeout:  time.Duration(timeout) * time.Minute}
+
 	r := mux.NewRouter()
 
 	postRouter := map[string]func(http.ResponseWriter, *http.Request){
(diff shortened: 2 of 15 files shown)


erikh:
ok, some general comments:

* Please wrap ALL rbd commands, not just the I/O ones.
* Wrap the pattern of setting the timeout in a function so the repetition is no longer necessary

volmaster/volmaster/cli.go
```diff
@@ -57,6 +57,11 @@ func main() {
 			Usage: "Set ttl of written locks; in seconds",
 			Value: 300,
 		},
+		cli.IntFlag{
+			Name:  "timeout",
+			Usage: "Set timeout for ceph commands; in seconds",
```
erikh: this needs to be in minutes

volsupervisor/volsupervisor/cli.go
```diff
@@ -46,6 +47,11 @@ func main() {
 			Usage: "URL for etcd",
 			Value: &cli.StringSlice{"http://localhost:2379"},
 		},
+		cli.IntFlag{
+			Name:  "timeout",
+			Usage: "Set timeout for ceph commands; in seconds",
```
erikh: again, minutes

erikh:
Also we need the mkfs calls rewritten too.

volmaster/daemon.go
```diff
@@ -95,7 +96,7 @@ func (d daemonConfig) handleRemove(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := removeVolume(vc); err != nil {
+	if err := removeVolume(vc, d.timeout); err != nil {
```
mapuri: this is minor but I am wondering if we should make `removeVolume()` and `createVolume()` `Daemon`'s methods. That way we don't need to pass daemon config around as args?

volsupervisor/snapshots.go
```diff
@@ -11,7 +11,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
-func wrapSnapshotAction(action func(config *config.TopLevelConfig, pool string, volume *config.VolumeConfig)) func(*volumeDispatch) {
+func wrapSnapshotAction(timeout time.Duration, action func(config *config.TopLevelConfig, pool string, volume *config.VolumeConfig, timeout time.Duration)) func(*volumeDispatch) {
```
mapuri: This is again minor and similar to previous comment may be we should make `wrapSnapshotAction` et-al `Daemon` methods to avoid having to pass around timeout as argument.

I realize supervisor doesn't yet have a config of it's own but may be we can introduce one?

volmaster/daemon.go
erikh: I'm not opposed to this, but now is probably out of scope; I'd like to do this with a separate patch. I'll put in a ticket (take it if you want to).

erikh: #132

mapuri: yes, sgtm @erikh

executor/executor.go
```diff
@@ -71,6 +71,21 @@ func New(cmd *exec.Cmd) *Executor {
 	}
 }
 
+// NewWithTimeout creates a new executor from an *exec.Cmd and a timeout.
+// You may modify the values before calling Start(). See Executor for
+// more information.
+func NewWithTimeout(cmd *exec.Cmd, timeout time.Duration) *Executor {
```
erikh: This could just be `New` with `SetTimeout` to avoid code duplication

volmaster/daemon.go
```diff
@@ -44,8 +45,8 @@ type volumeGet struct {
 }
 
 // Daemon initializes the daemon for use.
-func Daemon(config *config.TopLevelConfig, ttl int, debug bool, listen string) {
-	d := daemonConfig{config, ttl}
+func Daemon(config *config.TopLevelConfig, ttl, timeout int, debug bool, listen string) {
+	d := daemonConfig{config, ttl, time.Duration(timeout) * time.Minute}
```
erikh: can we use struct fields here? sorry, I know they weren't there before.

erikh:
LGTM, but wait for CI to merge please

PR #131, opened 2016-02-05
Fix docker

erikh:
This fixes docker 1.10 compatibility. It incorporates #129 and #127 which are already LGTM'd.

Please review immediately if you can; the build is broken without it. Just the last two commits are related to this change.

/cc @mapuri @unclejack

### DIFF
--- a/config/use.go
+++ b/config/use.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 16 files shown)


mapuri:
LGTM

do you know if docker plans to make `volumeGet`, `volume` and `volumeList` structures public? Otherwise seems like we need to manually keep them in sync.

erikh:
I've been told no before by cpuguy, but if it changes I will definitely update to take full advantage of it.

PR #134, opened 2016-02-05
Fix mapimage

erikh:
the ceph driver `mapImage` call was doing too much and causing trouble.


mapuri:
LGTM

erikh:
this now has a few docker 1.10 and race fixes in the pipe, please take another look @mapuri @unclejack

mapuri:
still LGTM

PR #138, opened 2016-02-10
Global configuration

### LINKED ISSUE #80: Global configuration in etcd
This would provide top-level configuration the volmaster, volsupervisor, and volplugin share and read (and refresh) from accordingly.

The system should be designed to:

* Handle dynamic updates (watch points and re-registration)
* volcli commands, e.g. `volcli config upload`
* handle at least these two fields for now:
  * TTL
  * Debug

See comments in #78 for more information

erikh:
closes #80.

Note that there are also two additional patches. They improve reliability throughout the system; were discovered while making the global config patch.

### DIFF
--- a/config/global.go
+++ b/config/global.go
@@ -1,0 +1,102 @@
+package config
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/coreos/etcd/client"
+	"golang.org/x/net/context"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// DefaultGlobalTTL is the TTL used when no TTL exists.
+const DefaultGlobalTTL = 30 * time.Second
+
+var (
+	timeoutFixBase = time.Minute
+	ttlFixBase     = time.Second
+)
+
+// Global is the global configuration.
+type Global struct {
+	Debug   bool
+	Timeout time.Duration
+	TTL     time.Duration
+}
+
+// PublishGlobal publishes the global configuration.
+func (tlc *TopLevelConfig) PublishGlobal(g *Global) error {
+	gcPath := tlc.prefixed("global-config")
+
+	value, err := json.Marshal(MultiplyGlobalParameters(g))
+	if err != nil {
+		return err
+	}
+
+	if _, err := tlc.etcdClient.Set(context.Background(), gcPath, string(value), &client.SetOptions{PrevExist: client.PrevIgnore}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetGlobal retrieves the global configuration.
+func (tlc *TopLevelConfig) GetGlobal() (*Global, error) {
+	resp, err := tlc.etcdClient.Get(context.Background(), tlc.prefixed("global-config"), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	global := &Global{}
+	if err := json.Unmarshal([]byte(resp.Node.Value), global); err != nil {
+		return nil, err
+	}
+
+	if global.TTL == 0 {
+		global.TTL = DefaultGlobalTTL
+	}
+
+	return global, nil
+}
+
+// MultiplyGlobalParameters multiplies the paramters by a fixed base to allow them to
+// be converted to proper time.Durations. This is to allow the global
+// configuration to have small numbers of second or minutes, instead of giant
+// numbers of nanoseconds in the JSON configuration.
+// See also: DivideGlobalParameters.
+func MultiplyGlobalParameters(global *Global) *Global {
+	global2 := *global
+	global2.TTL = ttlFixBase * global2.TTL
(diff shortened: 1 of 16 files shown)


unclejack:
LGTM

config/global.go
```diff
@@ -0,0 +1,95 @@
...
+
+var (
+	timeoutFixBase = time.Minute
+	ttlFixBase     = time.Second
+)
+
+// Global is the global configuration.
+type Global struct {
+	Debug   bool
+	Timeout time.Duration
```
mapuri: I think we may need to name these fields in a slightly more descriptive way as otherwise it will be hard to tell what are they used for in absence of a commanline help string, like perhaps this field should be called `CephTimeout` . Same will be true with ttl and may be we should call it `MasterAliveTTL` or similar?

volcli/commands.go
```diff
@@ -29,6 +29,27 @@ var GlobalFlags = []cli.Flag{
...
+			{
+				Name:        "upload",
+				ArgsUsage:   "accepts configuration from stdin",
+				Description: "Uploads a the global configuration to etcd. Accepts JSON.",
+				Usage:       "Upload a tenant to etcd",
+				Action:      GlobalUpload,
+			},
+			{
+				Name:        "get",
+				Flags:       VolmasterFlags,
```
mapuri: are these flags used?

volcli/volcli.go
```diff
@@ -37,6 +37,68 @@ func ppJSON(v interface{}) ([]byte, error) {
...
+	if err != nil {
+		errExit(ctx, err, false)
+	}
+
+	fmt.Println(string(content))
+}
+
+// GlobalUpload uploads the global configuration
+func GlobalUpload(ctx *cli.Context) {
+	if len(ctx.Args()) != 0 {
```
mapuri: Looks like we are expecting the config to come from stdin and we don't seems to be using the `args` anywhere in this function, so is this check needed?

volplugin/handlers.go
```diff
@@ -226,6 +227,7 @@ func mount(master, host string, ttl int) func(http.ResponseWriter, *http.Request
 
 		actualSize, err := volConfig.Options.ActualSize()
 		if err != nil {
+			removeStopChan(uc.Request.Name)
```
mapuri: I think we can `defer` the call to `removeStopChan()`?

mapuri:
LGTM other than a few minor comments and clarifications.

config/global.go
erikh: ok, will do.

volcli/commands.go
erikh: See line 46 of volcli.go

volcli/volcli.go
erikh: it ensures there are no additional arguments, but I can check if that's necessary.

volplugin/handlers.go
erikh: If we do that it gets cleaned up here too: https://github.com/erikh/volplugin/blob/global-configuration/volplugin/handlers.go#L261

the stop chan (and the heartbeat)n is held open while the volumes are mounted, so we can't do that.

volcli/commands.go
mapuri: ah, got it.

volcli/volcli.go
mapuri: makes sense. I think we can leave it as is as we are doing the same at other places.

volplugin/handlers.go
mapuri: oh, sorry I should have been more clear, I meant do something like below which will ensure we only cleanup in case of error. But current way works too, it's just seemed that we have to remember adding this call before every return :
```
defer func() {
    if err != nil {
        removeStopChan(uc.Request.Name)
    }
}()
```

erikh: The problem with this is that code expects err to be global scope, I really bit myself trying to do it this way last time in some of the rbd locking code. There's really no good pattern in go for this, I think. I mean, if you think you can get it right and not make the code harder to read, I'd love a patch, but I don't think I can do it.

mapuri: ah, makes sense, we can leave it as is then. Let me also think about this a bit more and will submit a patch if I find a similar pattern else where.

unclejack:
@mapuri PTAL

mapuri:
@unclejack 

LGTM overall. This comment https://github.com/contiv/volplugin/pull/138#discussion_r52943476 is not incorporated yet.

erikh:
do not merge this branch yet. it will get merged after the demo next week.

erikh:
@mapuri I think I'm going to leave these parameters alone for now; I will revisit them when we have confusing parameters. :)

mapuri:
cool, sgtm @erikh . changes LGTM

PR #140, opened 2016-02-10
ansible/roles/docker: use overlayfs storage backend.

erikh:
Merging as soon as tests pass.

This cuts runtime for tests in half.


unclejack:
LGTM

PR #143, opened 2016-02-19
volplugin,systemtests: ensure no FDs are carried by keepalive connections over the unix socket.

erikh:
/cc @shaleman @vvb 

This fixes fd leaks caused by API hyperactivity in docker 1.10.


vvb:
LGTM

PR #144, opened 2016-02-22
Stability enhancements

erikh:
This *removes* RBD locking and instead relies on our etcd locking strategy. RBD locks were in some situations unremovable, so I figured this was the best solution for now.

This also brings a few more enhancements to the tests while debugging this, and an upgrade to docker 1.10.1 in the ansible.


mapuri:
LGTM

PR #145, opened 2016-02-22
improve volcli errors

mapuri:
This change:
- improves the argument count related error string to be more explicit
- adds two List clis where the argument count check was missing to make it consistent with other places.

/cc @erikh @unclejack

### DIFF
--- a/volcli/volcli_test.go
+++ b/volcli/volcli_test.go
@@ -1,0 +1,129 @@
+package volcli
+
+import (
+	"flag"
+	. "testing"
+
+	"github.com/codegangsta/cli"
+
+	. "gopkg.in/check.v1"
+)
+
+type volcliSuite struct {
+}
+
+var _ = Suite(&volcliSuite{})
+
+func TestVolcli(t *T) { TestingT(t) }
+
+func (s *volcliSuite) TestErrorReturns(c *C) {
+	testMap := map[string]struct {
+		f    func(ctx *cli.Context) (bool, error)
+		args []string
+		err  error
+	}{
+		"tenantUpload": {
+			f:    tenantUpload,
+			args: []string{},
+			err:  errorInvalidArgCount(0, 1, []string{}),
+		},
+		"tenantDelete": {
+			f:    tenantDelete,
+			args: []string{},
+			err:  errorInvalidArgCount(0, 1, []string{}),
+		},
+		"tenantGet": {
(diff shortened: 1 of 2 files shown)


erikh:
I'm going to try this patch out by hand. I just want to see this in action is all, so give me a day or so to do that.

mapuri:
sgtm, thanks @erikh 

actually may be I can add unit-tests for the error strings let me look into that as well

mapuri:
@erikh I have updated the diff to add a few unit tests for error return paths. I had to refactor the functions a bit to make it easier to unit-test. PTAL and let me know if I didn't go nuts here :)

volcli/volcli_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type volcliSuite struct {
+}
+
+var _ = Suite(&volcliSuite{})
+
+func TestVolcli(t *T) { TestingT(t) }
+
+func (s *volcliSuite) SetUpTest(c *C) {
```
erikh: these shouldn't be necessary, you can remove them

erikh:
seeing an issue with `systemtestSuite.TestIntegratedRemoveWhileMount`. you can run:

`godep go test -v ./systemtests -check.v -check.f systemtestSuite.TestIntegratedRemoveWhileMount` to run it independently.

erikh:
I love this! Let's get it merged. First, I need to fix some things in the volplugin test runs though.

mapuri:
thanks @erikh . I have taken care of the review comment. I am trying to debug the system test failure but it seems unrelated to this commit.

erikh:
it is, disregard it, I'm working on it.

mapuri:
cool, thanks @erikh. I have squashed the commits, please feel free to merge as appropriate.

erikh:
the same test is failing, which makes me think there's a bug in this patch. this test calls `volcli` in a number of spots, can you validate that it is not broken by your patch?

mapuri:
@erikh 

I have identified the issue, basically in `volumeRemove()` when Post to volmaster is done and it is going to return the 500 status, that error was not translated to the volcli error. This got exposed now as I replaced `errExit()` calls with `return` instead.

 I have updated the patch with the fix and re-triggered the sanity. PTAL

erikh:
awesome thanks!

PR #146, opened 2016-02-22
storage/backend/ceph: refactor all `showmapped` calls so that they come from the same output.

erikh:


erikh:
I should point out that this limits our support for older ceph versions, but ensures no potential for mismatched output (and more importantly, it's json formatted so it's easier to read).

mapuri:
LGTM

PR #149, opened 2016-02-24
volplugin: guard against listen errors with a log message

erikh:


mapuri:
LGTM

PR #151, opened 2016-02-24
Test improvements

erikh:
Several commits related to improving test reliability and flexibility

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@ start: install-ansible
 	vagrant up
 	make build
 
+big:
+	BIG=1 vagrant up
+	make build
+
 stop:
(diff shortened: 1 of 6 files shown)


erikh:
can I get a review of this? @mapuri @unclejack

mapuri:
LGTM

just for my understanding, is `system-test-big` going to be the make target for stress and scale tests? Also you have not added to the `ci` target, is it because you see it run less often and in perhaps a separate jenkins job?

erikh:
I just want it in there right now; we can evaluate whether or not it's 
best for the CI later.

mapuri:
ok, sgtm

PR #152, opened 2016-02-26
Dm thin provisioning

erikh:
This enables DM thin provisioning for anyone who configures an available disk to `docker_device` which will enable the feature.

Disk performance is pretty dramatic so I recommend following the Vagrantfile to enable this in cluster and netplugin.


erikh:
You have to review independent commits since the merges are in there.

ansible/roles/docker/tasks/main.yml
```diff
@@ -38,29 +46,69 @@
...
   tags:
     - prebake-for-dev
 
+# XXX: this needs to happen twice after setting up the thin provisioner because
+#      of some docker bug I've not investigated.
+- name: restart docker (first time)
+  service:
+    name: docker
+    state: restarted
+  when:
```
mapuri: this condition needs to be appended with condition at https://github.com/contiv/volplugin/pull/152/files#diff-8a4c758b73a215ce668fd13badf3fb17R92 to continue to conditionally `restart docker` even when `thin_provisioned` is not defined

Also I need to test this but if `thin_provisioned` is not defined, this might fail so may be we will need to test that as well.

The condition could be:
```
((docker_service_state.stderr | match('.*docker.service changed on disk.*')) or
 (docker_service_state.stdout != 'Active: active')) or
(thin_provisioned | defined and
 thin_provisioned.rc == 0)
```

ansible/roles/docker/vars/main.yml
```diff
@@ -2,3 +2,4 @@
 # role variable for the swarm service
 
 docker_api_port: 2385
+thin_provisioned: false
```
mapuri: this file has been removed in `ansible/master` as part of fixing https://github.com/contiv/ansible/issues/61, we will need to move the new variable in `roles/docker/defaults/main.yml` file

mapuri:
@erikh will using this device make sure that the existing docker images in the packer box are preserved? I might have missed but does this go along @unclejack 's PR which tries to preserve the images https://github.com/contiv/build/pull/54

ansible/roles/docker/vars/main.yml
erikh: it's not a default though, it's a var. it doesn't leave the role and we 
don't want it set outside of the role.

erikh:
No it doesn't.

ansible/roles/docker/tasks/main.yml
erikh: docker is forcefully stopped above if the conditions are met, and then forcefully started in the step right after this one. I don't think this is necessary.

mapuri: > docker is forcefully stopped above if the conditions are met

hmm, may be I missed this, where is docker forcefully/unconditionally stopped? Unconditional stop/restart have side effects of killing running containers on replay of a playbook so we should avoid it as much possible.

I might not have been very clear earlier in my previous comment. My primary concern was that with the proposed check there is some mental deduction that needs to happen to be able to determine if this condition will do the right thing (like will it correctly re/start docker if the `docker_device` is not set?), by keeping the old condition in here ensures that it stays easy to deduce.

Also I did some testing with this patch to narrow down the problems, if any, with the proposed condition:
- this task fails if `thin_provisioned` is undefined but the error get's silently ignored as `ignore_errors` is set which I think is needed due to a docker bug but I guess we don't want the undefined variable error to be skipped.
- So I fixed that as follows and verified that it correctly get's skipped on calling `make provision`:
```
(thin_provisioned | default(false) | bool ) and
 thin_provisioned.rc == 0
```
- With the above fix, next I tried the scenario where `docker_device` is not set at time of doing `make start`. I noticed that this task is skipped i.e. docker is not at all started by this task, when I thought it should have started it.

- Changing the condition to following preserves the behavior of starting docker as today:
```
"((docker_service_state.stderr | match('.*docker.service changed on disk.*')) or
 (docker_service_state.stdout != 'Active: active')) or
((thin_provisioned | default(false) | bool) and
 thin_provisioned.rc == 0)"
```

ansible/roles/docker/vars/main.yml
mapuri: yes, I understand that vars can be defined at multiple places. But I started keeping all vars in `defaults` so there is just one way of introducing new vars. Otherwise it is hard to set a guideline for adding vars especially when a var is going to have a default value. I moved all vars to `defaults` for other reasons described in https://github.com/contiv/ansible/issues/61.

ansible/roles/docker/tasks/main.yml
```diff
@@ -38,29 +46,69 @@
...
+
 - name: copy systemd units for docker tcp socket settings
   template: src=docker-tcp.j2 dest=/etc/systemd/system/docker-tcp.socket
   register: docker_tcp_socket
 
   # tcp socket service requires docker service to be started after it
-- name: start docker tcp socket service
-  shell: sudo systemctl stop docker && sudo systemctl start docker-tcp.socket && sudo systemctl start docker
-  when: (docker_tcp_socket | changed) or not docker_tcp_socket_state.stdout | match("Active.* active")
+- name: reload and stop docker
```
mapuri: this task should perhaps be named `reload systemd configuration`

ansible/roles/docker/tasks/main.yml
```diff
@@ -38,29 +46,69 @@
...
+    - prebake-for-dev
 
 - name: remove the docker key file, if any. It shall be regenerated by docker on restart
   file: name=/etc/docker/key.json  state=absent
+  when: "(docker_service_state.stderr | match('.*docker.service changed on disk.*')) or (docker_service_state.stdout != 'Active: active')"
 
 # XXX: service module doesn't do daemon-reload yet, so need to use shell module here
 # https://github.com/ansible/ansible-modules-core/issues/191
-- name: restart docker
+- name: reload docker
```
mapuri: this task should perhaps be named `reload systemd configuration`

ansible/roles/docker/vars/main.yml
erikh: my next comment will explain this, but since we never actually *set* 
defaults, the variable has to be set somehow and it is not used outside 
the role. Therefore, I'm pretty sure `vars` is the best place for this, 
but I'm not going to push the issue.

Case in point: I don't want anyone to override this ever.

ansible/roles/docker/tasks/main.yml
erikh: On 27 Feb 2016, at 15:57, Madhav Puri wrote:

>> tags:
>>   - prebake-for-dev
>>
>> +# XXX: this needs to happen twice after setting up the thin 
>> provisioner because
>> +#      of some docker bug I've not investigated.
>> +- name: restart docker (first time)
>> +  service:
>> +    name: docker
>> +    state: restarted
>> +  when:
>
>> docker is forcefully stopped above if the conditions are met
>
> hmm, may be I missed this, where is docker forcefully/unconditionally 
> stopped? Unconditional stop/restart have side effects of killing 
> running containers on replay of a playbook so we should avoid it as 
> much possible.

https://github.com/erikh/volplugin/commit/be34e26cef9c29ca8e6338fa6c4c9d6ed937b560#diff-8a4c758b73a215ce668fd13badf3fb17R60

explicitly, not unconditionally.

> I might not have been very clear earlier in my previous comment. My 
> primary concern was that with the proposed check there is some mental 
> deduction that needs to happen to be able to determine if this 
> condition will do the right thing (like will it correctly re/start 
> docker if the `docker_device` is not set?), by keeping the old 
> condition in here ensures that it stays easy to deduce.

I've already verified this, believe me. I tested in the following 
conditions on both vmware and virtualbox:

* without thin provisioning
* with thin provisioning
* second ansible run on thin provisioning (both systems)

> Also I did some testing with this patch to narrow down the problems, 
> if any, with the proposed condition:
> - this task fails if `thin_provisioned` is undefined but the error 
> get's silently ignored as `ignore_errors` is set which I think is 
> needed due to a docker bug but I guess we don't want the undefined 
> variable error to be skipped.
> - So I fixed that as follows and verified that it correctly get's 
> skipped on calling `make provision`:
> ```
> (thin_provisioned | default(false) | bool ) and
> thin_provisioned.rc == 0
> ```
> - With the above fix, next I tried the scenario where `docker_device` 
> is not set at time of doing `make start`. I noticed that this task is 
> skipped i.e. docker is not at all started by this task, when I thought 
> it should have started it.

No, if the var doesn't go away like you keep suggesting, this never 
happens. It's impossible for it to happen, because the var is set to 
false as soon as the role loads.

This code works.

-Erik

ansible/roles/docker/tasks/main.yml
erikh: Aye, I will fix.

ansible/roles/docker/vars/main.yml
mapuri: sure, we can leave this in vars. I think we need to cleanup `docker_api_port: 2385` definition then, I think it has slipped in due to merge.

ansible/roles/docker/tasks/main.yml
mapuri: > No, if the var doesn't go away like you keep suggesting, this never happens.

hmm, yes, you are right the var should have been defined. I looked at the logs again (`make provision` case) and this is what I saw which I assumed was coming from this variable being undefined but I was mistaken. I think there is still something that might need to be fixed in the condition but I might be missing something as you didn't see it.

```
TASK [docker : restart docker (first time)] ************************************
fatal: [mon0]: FAILED! => {"failed": true, "msg": "The conditional check 'thin_provisioned and thin_provisioned.rc == 0' failed. The error was: error while evaluating conditional (thin_provisioned and thin_provisioned.rc == 0): 'dict object' has no attribute 'rc'\n\nThe error appears to have been in '/home/ladmin/goroot/src/github.com/contiv/volplugin/ansible/roles/docker/tasks/main.yml': line 102, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n#      of some docker bug I've not investigated.\n- name: restart docker (first time)\n  ^ here\n"}
...ignoring
TASK [docker : ensure docker is started] ***************************************
changed: [mon0]
```

mapuri: > explicitly, not unconditionally

ah, I see but note that this is topped only when tcp-socket service is setup for first time i.e. this condition will not happen on subsequent re-run of the playbook as tcp-service would already have been setup.

ansible/roles/docker/vars/main.yml
erikh: will do

ansible/roles/docker/tasks/main.yml
erikh: Yup

erikh: Whoops! Ok, that's a bug.

-Erik

unclejack:
In spite of what GitHub is saying, this doesn't merge cleanly on the CLI.

erikh:
ok, @mapuri @unclejack can you take a final pass at this?

ansible/roles/docker/defaults/main.yml
```diff
@@ -2,3 +2,6 @@
 # Default values for docker role
 
 docker_version: 1.9.1
+docker_api_port: 2375
```
mapuri: @erikh this needs to be 2385. 2375 is swarm port in our ansible.

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
   tags:
     - prebake-for-dev
 
+# XXX: this needs to happen twice after setting up the thin provisioner because
+#      of some docker bug I've not investigated.
+- name: restart docker (first time)
+  service:
+    name: docker
+    state: restarted
+  when: thin_provisioned
```
mapuri: @erikh this restarts docker on `make provision`, that will stop any running container. I think that is due to the fact that `thin_provisioned` is not set to `false` (as when the `lvcreate contiv-dockerthin` task is skipped it still ends up register this variable).

here is what I see, note that the `restart docker (first time)` task in both `make start` and `make provision` is triggered causing docker to be restarted kind of unconditionally:

## make start ##
```
TASK [docker : copy systemd units for docker tcp socket settings] **************
changed: [mon0]
changed: [mon1]
changed: [mon2]

TASK [docker : reload systemd configuration] ***********************************
changed: [mon0]
changed: [mon1]
changed: [mon2]

TASK [docker : stop docker] ****************************************************
ok: [mon0]
ok: [mon1]
ok: [mon2]

TASK [docker : start docker-tcp service] ***************************************
changed: [mon0]
changed: [mon1]
changed: [mon2]

TASK [docker : check docker service state] *************************************
changed: [mon2]
changed: [mon1]
changed: [mon0]

TASK [docker : remove the docker key file, if any. It shall be regenerated by docker on restart] ***
changed: [mon1]
changed: [mon0]
changed: [mon2]

TASK [docker : reload docker systemd configuration] ****************************
changed: [mon0]
changed: [mon1]
changed: [mon2]

TASK [docker : restart docker (first time)] ************************************
fatal: [mon1]: FAILED! => {"changed": false, "failed": true, "msg": "Warning: Stopping docker.service, but it can still be activated by:\n  docker-tcp.socket\nJob for docker.service failed because the control process exited with error code. See \"systemctl status docker.service\" and \"journalctl -xe\" for details.\n"}
...ignoring
fatal: [mon2]: FAILED! => {"changed": false, "failed": true, "msg": "Warning: Stopping docker.service, but it can still be activated by:\n  docker-tcp.socket\nJob for docker.service failed because the control process exited with error code. See \"systemctl status docker.service\" and \"journalctl -xe\" for details.\n"}
...ignoring
fatal: [mon0]: FAILED! => {"changed": false, "failed": true, "msg": "Warning: Stopping docker.service, but it can still be activated by:\n  docker-tcp.socket\nJob for docker.service failed because the control process exited with error code. See \"systemctl status docker.service\" and \"journalctl -xe\" for details.\n"}
...ignoring

TASK [docker : ensure docker is started] ***************************************
changed: [mon1]
changed: [mon0]
changed: [mon2]
```

## make provision ##
```
TASK [docker : lvcreate contiv-dockerthin] *************************************
skipping: [mon0]
skipping: [mon1]
skipping: [mon2]

TASK [docker : copy systemd units for docker tcp socket settings] **************
ok: [mon2]
ok: [mon1]
ok: [mon0]

TASK [docker : reload systemd configuration] ***********************************
changed: [mon1]
 [WARNING]: Consider using 'become', 'become_method', and 'become_user' rather
than running sudo

changed: [mon2]
changed: [mon0]

TASK [docker : stop docker] ****************************************************
ok: [mon1]
ok: [mon0]
ok: [mon2]

TASK [docker : start docker-tcp service] ***************************************
ok: [mon0]
ok: [mon1]
ok: [mon2]

TASK [docker : check docker service state] *************************************
fatal: [mon1]: FAILED! => {"changed": true, "cmd": "sudo systemctl status docker | grep 'Active.*active' -o", "delta": "0:00:00.020530", "end": "2016-03-01 12:07:15.903307", "failed": true, "rc": 1, "start": "2016-03-01 12:07:15.882777", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": ["Consider using 'become', 'become_method', and 'become_user' rather than running sudo"]}
...ignoring
fatal: [mon0]: FAILED! => {"changed": true, "cmd": "sudo systemctl status docker | grep 'Active.*active' -o", "delta": "0:00:00.021246", "end": "2016-03-01 12:07:15.921782", "failed": true, "rc": 1, "start": "2016-03-01 12:07:15.900536", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": ["Consider using 'become', 'become_method', and 'become_user' rather than running sudo"]}
...ignoring
fatal: [mon2]: FAILED! => {"changed": true, "cmd": "sudo systemctl status docker | grep 'Active.*active' -o", "delta": "0:00:00.032452", "end": "2016-03-01 12:07:15.957005", "failed": true, "rc": 1, "start": "2016-03-01 12:07:15.924553", "stderr": "", "stdout": "", "stdout_lines": [], "warnings": ["Consider using 'become', 'become_method', and 'become_user' rather than running sudo"]}
...ignoring

TASK [docker : remove the docker key file, if any. It shall be regenerated by docker on restart] ***
changed: [mon0]
changed: [mon1]
changed: [mon2]

TASK [docker : reload docker systemd configuration] ****************************
changed: [mon1]
changed: [mon0]
changed: [mon2]

TASK [docker : restart docker (first time)] ************************************
changed: [mon1]
changed: [mon0]
changed: [mon2]

TASK [docker : ensure docker is started] ***************************************
ok: [mon0]
ok: [mon2]
ok: [mon1]

```

erikh: ugh! I guess I missed this case. Thanks, will fix tonight.

erikh:
ok, one more look, I think you will be happy.

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
+
+- name: check docker service state
+  shell: sudo systemctl status docker | grep 'Active.*active' -o
+  ignore_errors: true
+  register: docker_service_state
+  tags:
+    - prebake-for-dev
 
 - name: remove the docker key file, if any. It shall be regenerated by docker on restart
   file: name=/etc/docker/key.json  state=absent
```
mapuri: this might be another merge issue, this task is run conditionally in ansible/master

https://github.com/contiv/ansible/blob/master/roles/docker/tasks/main.yml#L68

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
   #service: name=docker state=restarted
-  shell: sudo systemctl daemon-reload && sudo systemctl restart docker
+  shell: sudo systemctl daemon-reload
+  when: "(docker_service_state.stderr | match('.*docker.service changed on disk.*')) or (docker_service_state.stdout != 'Active: active')"
   tags:
     - prebake-for-dev
 
+# XXX: this needs to happen twice after setting up the thin provisioner because
+#      of some docker bug I've not investigated.
+- name: restart docker (first time)
```
mapuri: we need to add `prebake-for-dev` tag for this task for packer builds

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
+  service:
+    name: docker
+    state: restarted
+  when: thin_provisioned|changed
```
mapuri: thanks @erikh this works as expected now in `make start` and `make provision` cases when `docker_device` is **set**. However, I still see following failure on `make start` when `docker_device` is **not set**, may be defining `thin_provisioned` as a empty dictionary will fix this. Also I think in addition we will need to augment the condition with this condition https://github.com/contiv/volplugin/pull/152/files#diff-8a4c758b73a215ce668fd13badf3fb17R84 for the task to restart docker only if if it is not running.

**make start** failure logs:
```
TASK [docker : restart docker (first time)] ************************************
fatal: [mon1]: FAILED! => {"failed": true, "msg": "The conditional check 'thin_provisioned|changed' failed. The error was: |changed expects a dictionary\n\nThe error appears to have been in '/home/ladmin/goroot/src/github.com/contiv/volplugin/ansible/roles/docker/tasks/main.yml': line 90, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n#      of some docker bug I've not investigated.\n- name: restart docker (first time)\n  ^ here\n"}                                                       ...ignoring
fatal: [mon0]: FAILED! => {"failed": true, "msg": "The conditional check 'thin_provisioned|changed' failed. The error was: |changed expects a dictionary\n\nThe error appears to have been in '/home/ladmin/goroot/src/github.com/contiv/volplugin/ansible/roles/docker/tasks/main.yml': line 90, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n#      of some docker bug I've not investigated.\n- name: restart docker (first time)\n  ^ here\n"}                                                       ...ignoring
fatal: [mon2]: FAILED! => {"failed": true, "msg": "The conditional check 'thin_provisioned|changed' failed. The error was: |changed expects a dictionary\n\nThe error appears to have been in '/home/ladmin/goroot/src/github.com/contiv/volplugin/ansible/roles/docker/tasks/main.yml': line 90, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n#      of some docker bug I've not investigated.\n- name: restart docker (first time)\n  ^ here\n"}                                                       ...ignoring

TASK [docker : ensure docker is started] ***************************************
changed: [mon0]
changed: [mon1]
changed: [mon2]
```

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
+  ignore_errors: true
+
+- name: ensure docker is started
```
mapuri: we need to add `prebake-for-dev` tag for this task for packer builds

ansible/roles/docker/tasks/main.yml
```diff
@@ -41,26 +42,63 @@
...
+  service:
+    name: docker
+    state: started
```
mapuri: @erikh 

Now I am noticing that on `make provision` this task is toggling docker state (Verified from journalctl logs on one of the vms). I suspect this could be ansible module issue but may be we should put this under the above conditional too. wdyt?

```
TASK [docker : restart docker (first time)] ************************************
skipping: [mon0]
skipping: [mon1]
skipping: [mon2]

TASK [docker : ensure docker is started] ***************************************
changed: [mon1]
changed: [mon2]
changed: [mon0]
```

erikh: sure, let me look into it.

erikh: this won't work, and shouldn't change (I have no idea why it's changing). `started` also means to enable it on boot, so that might be the reason.

Either way I don't think this should be changed, and if there is a problem it is elsewhere in the ansible. I think I saw an issue on redhat where it's reinstalling docker each time.

Can we merge this?

ansible/roles/docker/tasks/main.yml
erikh: No, we actually *do* want to restart it if it is running, but only when thin provisioning has happened, so it picks up the new device and starts using it.

ansible/roles/docker/tasks/main.yml
erikh: I haven't imported latest ansible in this branch, deliberately.

erikh:
prebake-for-dev tag added, but I think that's all that's needed for now.

ansible/roles/docker/tasks/main.yml
mapuri: @erikh 

I didn't understand. This will create a merge conflict when we merge this to ansible/master, right? Or are you planning to remove the existing conditional? I think it will break things.

ansible/roles/docker/tasks/main.yml
mapuri: hmm, I think the error message I pasted is causing the task to be not run (it fails and failure is ignored) so restart is actually not triggered. We should fix that I think.

ansible/roles/docker/tasks/main.yml
mapuri: @erikh 

> started also means to enable it on boot

for service to be enabled we need to set the `enabled` attribute in `service` module. But I am ok if you want to leave this task as is.

erikh:
@mapuri PTAL and let me know if I can merge this

mapuri:
sure @erikh sorry to be nagging on this for too long. We can merge this now and carry the discussion on remaining comments later in the PR to merge to ansible repo. Or address them later as needed.

erikh:
@mapuri don't sweat it. is there something else I should fix right now? I think the other issues that I have not resolved are because of other issues.

ansible/roles/docker/tasks/main.yml
erikh: oh. you're right. hm. I will look at this asap.

erikh:
I'm going to merge this, let's revisit offline for better working of the docker role in general; I think it needs an overhaul.

mapuri:
cool, btw you will need to push this to ansible repo as well :)

erikh:
yeah I plan to push an amalgamated PR from volplugin since several PRs 
made changes

PR #154, opened 2016-02-27
Daemon structs

### LINKED ISSUE #132: Refactor: volmaster and volsupervisor need to take daemon structs as self
Right now we pass these around the `daemon` object quite a bit, and in volsupervisor's case, a lot of arguments that travel down the stack. We should organize these into structs the methods live on, so that they can re-use the same data without having ugly call sites.

erikh:
Fixes #132 

This patch implements daemon-wide structs for use by the services, to keep globals and so on. it cleans up the code a little.

### DIFF
--- a/volmaster/daemon.go
+++ b/volmaster/daemon.go
@@ -19,10 +19,11 @@ import (
 	"github.com/gorilla/mux"
 )
 
-type daemonConfig struct {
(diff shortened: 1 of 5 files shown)


mapuri:
LGTM

unclejack:
LGTM

PR #157, opened 2016-02-27
config: clean up units handling, add another test for 1GB to test powers of 1024 are being used.

erikh:


unclejack:
LGTM

PR #158, opened 2016-02-27
Metacharacter management

### LINKED ISSUE #75: define and restrict a set of metacharacters in volume names
Right now `/` and `.` are special characters. There's bound to be a few more as we implement data stores. However, we do not ensure that the volume requested to be created does not have these characters in them already. Since we do a lot of transliteration of these two (ceph does not accept `/` in volume names) it's important we do not kill ourselves in the process.

This should be easy to do.

erikh:
These patches push the management of metacharacters other than `/` down to the storage driver so they can deal with it.

fixes #75

### DIFF
--- a/storage/backend/ceph/ceph_test.go
+++ b/storage/backend/ceph/ceph_test.go
@@ -199,3 +199,25 @@ func (s *cephSuite) TestMounted(c *C) {
 	c.Assert(driver.Unmount(driverOpts), IsNil)
 	c.Assert(driver.Destroy(driverOpts), IsNil)
 }
+
+func (s *cephSuite) TestInternalNames(c *C) {
+	driver := NewDriver()
+	out, err := driver.InternalName("tenant1/test")
+	c.Assert(err, IsNil)
+	c.Assert(out, Equals, "tenant1.test")
+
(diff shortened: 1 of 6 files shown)


storage/backend/ceph/ceph_test.go
```diff
@@ -199,3 +199,21 @@ func (s *cephSuite) TestMounted(c *C) {
 	c.Assert(driver.Unmount(driverOpts), IsNil)
 	c.Assert(driver.Destroy(driverOpts), IsNil)
 }
+
+func (s *cephSuite) TestInternalNames(c *C) {
+	driver := NewDriver()
+	out, err := driver.InternalName("tenant1/test")
```
mapuri: @erikh Is 'tenant1/test/foo' an allowed format? From #75 looks like ceph won't allow creating a volume with `/` so just wondering if we should catch it here or let ceph error out.

erikh: yeah hm you're right.

erikh:
@mapuri @unclejack PTAL, squashed and ready for merge (presuming the overlay bug doesn't hit in CI)

mapuri:
LGTM

PR #162, opened 2016-03-03
Add a null/no-op volume driver

### LINKED ISSUE #153: develop a null volume driver
This driver shall be primarily used for testing and benchmarking voplugin functionality in isolation.

mapuri:
The changes here introduce a null driver that basically doesn't perform any actual disk operations.

The driver to use for a volume is set in `volume-options` using field called `backend`. The default backend is `ceph` and a user can override it by specifying the `backend` field in `default-options` structure.

The primary use case of the null driver will be to use it in scale/battery tests to exercise/benchmark volplugin alone  and I plan to work on that next. But I put out this PR to make sure I am working in right direction and also to keep the patch small.

fixes #153 

@erikh @unclejack PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ node_modules
 dist
 docs/docs
 .DS_Store
+*.swp
+cscope.*
--- a/systemtests/util_test.go
+++ b/systemtests/util_test.go
@@ -11,6 +11,7 @@ import (
 	utils "github.com/contiv/systemtests-utils"
 	"github.com/contiv/vagrantssh"
 	"github.com/contiv/volplugin/config"
+	"github.com/contiv/volplugin/storage/backend/ceph"
 )
 
 func (s *systemtestSuite) mon0cmd(command string) (string, error) {
@@ -31,7 +32,7 @@ func (s *systemtestSuite) readIntent(fn string) (*config.PolicyConfig, error) {
 		return nil, err
 	}
 
-	cfg := &config.PolicyConfig{}
+	cfg := config.NewPolicyConfig(ceph.BackendName)
 
 	if err := json.Unmarshal(content, cfg); err != nil {
 		return nil, err
@@ -146,7 +147,8 @@ func (s *systemtestSuite) rebootstrap() error {
 		return err
 	}
 
-	if _, err := s.uploadIntent("policy1", "intent1"); err != nil {
+	if out, err := s.uploadIntent("policy1", "intent1"); err != nil {
+		log.Errorf("Intent upload failed. Error: %v, Output: %s", err, out)
 		return err
 	}
 
--- a/volcli/volcli.go
+++ b/volcli/volcli.go
@@ -53,64 +53,86 @@ func ppJSON(v interface{}) ([]byte, error) {
 
 // GlobalGet retrives the global configuration and displays it on standard output.
 func GlobalGet(ctx *cli.Context) {
-	if len(ctx.Args()) != 0 {
-		errExit(ctx, fmt.Errorf("Invalid arguments"), true)
-	}
+	execCliAndExit(ctx, globalGet)
+}
 
+func queryGlobalConfig(ctx *cli.Context) (*config.Global, error) {
 	resp, err := http.Get(fmt.Sprintf("http://%s/global", ctx.String("volmaster")))
 	if err != nil {
-		errExit(ctx, err, false)
+		return nil, err
 	}
 
 	content, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		errExit(ctx, err, false)
+		return nil, err
 	}
 
 	if resp.StatusCode != 200 {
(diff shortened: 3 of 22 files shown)


systemtests/util_test.go
```diff
@@ -31,7 +32,11 @@ func (s *systemtestSuite) readIntent(fn string) (*config.TenantConfig, error) {
 		return nil, err
 	}
 
-	cfg := &config.TenantConfig{}
+	cfg := &config.TenantConfig{
```
erikh: since these are only used for tests, let's change the testdata instead of adding this code.

volcli/volcli.go
```diff
@@ -133,7 +134,11 @@ func tenantUpload(ctx *cli.Context) (bool, error) {
 		return false, err
 	}
 
-	tenant := &config.TenantConfig{}
+	tenant := &config.TenantConfig{
```
erikh: there's a lot of repetition about this. Let's instead make  amethod, `config.NewTenantConfig(driver string)` that performs this construction. This can also be used in the tests where you do this.

Does this work for you?

erikh:
Final comment: I would like to see the notion of a driver owning its own name -- can we add a `Name() string` method to the interface and each of the drivers that returns the right thing? It could be useful for diagnostic purposes later.

erikh:
Oh. The notion of a global default would be very nice, if you have the time to incorporate that.

systemtests/util_test.go
mapuri: yes, we can do this.

Or see my next comment, may using `config.NewDefaultTenantConfig()` here as well might be better as that will help us test that `ceph` is picked as default driver when one is not specified by the user.

volcli/volcli.go
mapuri: yeah, I agree the repetition can be avoided by using a function for default config.

I am wondering if having `config.NewDefaultTenantConfig()` method will be better which will return an empty config with `backend` set to `ceph`. Does it sound right?

Later we can may be add other things to this method as well like I saw in volcli tests we by-default add a `"ext4": "mkfs.ext4 -m0 %"` filesystem to the config options. But I might be wrong about it's purpose.

mapuri:
thanks for the review @erikh 

> can we add a Name() string method to the interface and each of the drivers that returns the right thing?

that is a good idea, let me try it out.

> The notion of a global default would be very nice, if you have the time to incorporate that.

I am not sure if I completely understood this. By global-default do you imply having ceph as a default backed. If so I am wondering if `config.NewDefaultTenantConfig()` function as suggested above will address this?

Let me work on a patch with this change in a bit as it might be more clear that way.

erikh:
@mapuri I mean a global default as in the new "global configuration" (see examples in `systemtests/testdata/global1.json`, which would revert to ceph if unspecified.

erikh:
and yes, a perfect place to consume this would be in your new function.

mapuri:
@erikh this should be ready for another look. Following is the highlight of changes:

- Added `config.NewTenantConfig` function to return a tenant config with specified backend name
- Added `config.NewGlobalConfig` function to return a global config with default backend name and ttl
- In tenant upload volcli use the backend name from global-config as a default when it is not provided by the user.
 - also fixed global config related volcli handlers to be consistent with other cli handlers
- added `Name()` method to `storage.Driver` interface which can be used to return the backend name of a gvien instance.

erikh:
LGTM. Thanks for the extra work on globals.

erikh:
let's wait for @unclejack to peek at it before merging.

erikh:
I'm going to merge #163 before this, are you comfortable rebasing against this @mapuri ?

.gitignore
```diff
@@ -13,3 +13,5 @@ node_modules
 dist
 docs/docs
 .DS_Store
+*.swp
+cscope.*
```
unclejack: @mapuri This change is unrelated to the commits.

unclejack:
LGTM with one nit to address: the extraneous change made to `.gitignore`.

.gitignore
mapuri: @unclejack yes, you are right. It is needed to avoid showing my ide and tags files in git status. I just slipped it in with this PR to avoid having to create a separate issue. Let me remove it and create a separate PR.

mapuri:
@erikh yes, should be fine. I will rebase once you have merged the other PR.

.gitignore
erikh: I dunno if this is necessary, let's just move forward with this patch.

Rebase is already going to be hell. :)

erikh:
@mapuri rebase please :)

mapuri:
yep, will take care of this in a bit :)

mapuri:
@erikh I have rebased the patch. PTAL when you get a chance.

mapuri:
uh..oh, don't bother, looks like it was already merged :)

PR #163, opened 2016-03-03
config,volplugin,volcli,volmaster: s/tenant/Policy/g

erikh:
Don't waste a lot of time reviewing this, it's just a sed job. Maybe the comments or help messages would be the thing to focus on, everything should pass on tests.


mapuri:
lgtm

Just curious, do we need a notion of tenant in volplugin eventually? May be it will be a policy-group or policy namespace of some sort.

erikh:
Maybe, not sure that policies don't serve that now the way they're architected, though.

mapuri:
sgtm. Also I think we should rename `config/tenant*.go` files to `config/policy*.go`?

erikh:
yeah we should! I will fix.

PR #170, opened 2016-03-06
Locking client + volplugin cleanup

erikh:
This isolates the code from volplugin that communicates with the volmaster for clarity and later refactoring purposes.

### DIFF
--- a/lock/client/client.go
+++ b/lock/client/client.go
@@ -1,0 +1,136 @@
+// Package client implements a client to the volmaster to acquire locks.
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/contiv/volplugin/config"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// Driver is the main force behind the lock module, it controls all methods and
+// several variables.
+type Driver struct {
+	master         string
+	mountStopChans map[string]chan struct{}
+	mountMutex     sync.Mutex
+}
+
+// NewDriver constructs a new driver.
+func NewDriver(master string) *Driver {
+	return &Driver{
+		master:         master,
+		mountStopChans: map[string]chan struct{}{},
+		mountMutex:     sync.Mutex{},
+	}
+}
+
+// AddStopChan adds a stop channel to a map for heartbeat tracking purposes.
+func (d *Driver) AddStopChan(name string) chan struct{} {
+	d.mountMutex.Lock()
+	defer d.mountMutex.Unlock()
+
+	stopChan := make(chan struct{})
+
+	if sc, ok := d.mountStopChans[name]; ok {
+		sc <- struct{}{}
+	}
+
+	d.mountStopChans[name] = stopChan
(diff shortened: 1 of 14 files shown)


erikh:
I've also added porting for the volmaster too.

lock/client/client.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/contiv/volplugin/config"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// Driver is the main force behind the lock module, it controls all methods and
+// several variables.
+type Driver struct {
+	master         string
+	mountStopChans map[string]chan struct{}
```
mapuri: I might have missed this, but I am not very clear on advantage of using a `struct{}` vs say `interface{}` or `bool` here. We don't seem to be interpreting/using the `struct` received on this channel anywhere, right?

lock/client/client.go
```diff
@@ -0,0 +1,136 @@
...
+	mountMutex     sync.Mutex
```
mapuri: minor: just wondering, if we shall just embed the `sync.Mutex` directly here? That way we can just call `d.Lock()` and `d.Unlock()` every, than more verbose `d.mountMutex...`

lock/client/client.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// NewDriver constructs a new driver.
+func NewDriver(master string) *Driver {
+	return &Driver{
+		master:         master,
+		mountStopChans: map[string]chan struct{}{},
```
mapuri: minor: I think `make( map[string]chan struct{})` might be more Go-like instantiation of an empty map?

lock/lock.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// NewDriver creates a Driver. Requires a configured TopLevelConfig.
+func NewDriver(config *config.TopLevelConfig, global *config.Global) *Driver {
+	return &Driver{Config: config, Global: global}
+}
+
+// ExecuteWithUseLock executes a function within a lock/context of the passed
+// *config.UseMount.
+func (d *Driver) ExecuteWithUseLock(uc *config.UseMount, runFunc func(d *Driver, uc *config.UseMount) error) error {
```
mapuri: I think `config.UseMount` in the method args should be replaced with `config.UseLocker`, right?

mapuri:
overall LGTM, just a few minor comments.

lock/client/client.go
erikh: struct{} is free: http://dave.cheney.net/2014/03/25/the-empty-struct

lock/client/client.go
erikh: perhaps, but I really want a better reason to rewrite the patch than 
this. :) Maybe in a future refactor.

lock/lock.go
erikh: yeah, you're right.

lock/client/client.go
mapuri: ah nice, didn't know about this, learned something new :) :+1:

lock/client/client.go
mapuri: sgtm :)

PR #171, opened 2016-03-07
Mountpoint global

erikh:
This provides the base mount point (hard-coded as a default previously as `/mnt/ceph`) as a configurable global.

I made some tests but I'm sorry for the size of this patch; it just touched everything.

Be aware this also depends on the #170 patch getting merged first.

### DIFF
--- a/config/global.go
+++ b/config/global.go
@@ -15,23 +15,26 @@ import (
 const DefaultGlobalTTL = 30 * time.Second
 
 var (
-	timeoutFixBase = time.Minute
-	ttlFixBase     = time.Second
+	timeoutFixBase   = time.Minute
+	ttlFixBase       = time.Second
+	defaultMountPath = "/mnt/ceph"
 )
 
 // Global is the global configuration.
 type Global struct {
-	Debug   bool
-	Timeout time.Duration
-	TTL     time.Duration
-	Backend string
+	Debug     bool
+	Timeout   time.Duration
+	TTL       time.Duration
+	Backend   string
+	MountPath string
 }
 
 // NewGlobalConfig returns global config with preset defaults
 func NewGlobalConfig() *Global {
 	return &Global{
-		TTL:     DefaultGlobalTTL,
-		Backend: ceph.BackendName,
+		TTL:       DefaultGlobalTTL,
+		Backend:   ceph.BackendName,
+		MountPath: defaultMountPath,
 	}
 }
 
@@ -63,7 +66,19 @@ func (tlc *TopLevelConfig) GetGlobal() (*Global, error) {
 		return nil, err
 	}
 
+	global.fixupParameters()
+
 	return global, nil
+}
+
+func (global *Global) fixupParameters() {
+	if global.TTL == 0 {
+		global.TTL = DefaultGlobalTTL
+	}
+
+	if global.MountPath == "" {
+		global.MountPath = defaultMountPath
+	}
 }
 
 // MultiplyGlobalParameters multiplies the paramters by a fixed base to allow them to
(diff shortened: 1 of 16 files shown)


erikh:
@mapuri @unclejack ping, PTAL

config/global.go
```diff
@@ -63,6 +66,14 @@ func (tlc *TopLevelConfig) GetGlobal() (*Global, error) {
 		return nil, err
 	}
 
+	if global.TTL == 0 {
```
mapuri: I think we can safely skip setting defaults here again as `NewGlobalConfig()` already sets the defaults and `json.Unmarshal` will not touch these values if they are not set.

lock/client/client.go
```diff
@@ -85,8 +85,6 @@ func (d *Driver) reportMountEndpoint(endpoint string, ut *config.UseMount) error
 		return err
 	}
 
-	fmt.Printf("%#v\n", d)
```
mapuri: looks like this is a left over print or may be this should be `log.Debugf()` ?

volplugin/volplugin.go
```diff
@@ -166,50 +166,56 @@ func (dc *DaemonConfig) getGlobal() {
...
 }
 
 func (dc *DaemonConfig) updateMounts() error {
-	cd := ceph.NewDriver()
-	mounts, err := cd.Mounted(dc.Global.Timeout)
-	if err != nil {
-		return err
-	}
+	for driverName := range backend.Drivers {
+		cd, err := backend.NewDriver(driverName, dc.Global.MountPath)
```
mapuri: @erikh I am just wondering if having a global mount-path work across drivers. Won't separate drivers step over each other, if they try to manage the same mount path?

lock/client/client.go
erikh: It's a removal; a spurious printf. :)

volplugin/volplugin.go
erikh: i don't think so. Policies are uniquely named, namespacing volumes. If 
the drivers manage themselves this way, there will be no conflicts.

config/global.go
erikh: Ah. I wondered about that, was being defensive. Ok, I will remove them 
and re-run the tests.

volplugin/volplugin.go
mapuri: ah, ok..got it.

I saw his code https://github.com/contiv/volplugin/pull/171/files#diff-1094dbd71a8251a1917ea1cbdff6520cR176 which made me think every driver might want to do this or do it differently which made me wonder if it was ok to have a global mountpath. But looking further down that code I see that the volumepath are namespaced, so as long as each driver just restricts to their owned volumes they should be fine.

lock/client/client.go
mapuri: oh, well yes you are right :) ... sorry for the spam.

mapuri:
LGTM

erikh:
This won't pass the CI: fwiw, this actually does fail because parameters 
are missing during json.Unmarshal time

mapuri:
@erikh 

> this actually does fail because parameters are missing during json.Unmarshal time

Isn't json.Unmarshal called before the checks? How do checks the fix it? Or were you referring to Unmarshal call in some test?

erikh:
look at the latest diff -- they search for default values and replace 
them with the actual defaults.

mapuri:
ok, I see. So the intent is to reset even user specified empty or zero values with the daemon defaults.

erikh:
yes, but a side effect; if someone sets either of those to an empty 
value, badness will happen.

PR #173, opened 2016-03-09
Watch

erikh:
This provides a high-level interface to etcd watches and ports WatchGlobal to use it.

Will be used by an upcoming volsupervisor rewrite.


mapuri:
LGTM

PR #175, opened 2016-03-10
Bump box

erikh:
This bumps the box version to 0.3.0 and adds improvements in provisioning speed.


erikh:
that's the sshd timing out because the code is hitting it too hard. I'll re-run, and address that in another patch.

PR #176, opened 2016-03-14
volsupervisor: rewritten

### LINKED ISSUE #156: snapshots race volume remove
I noticed today that the process of taking a snapshot can race `volcli volume remove` when the snapshot delay is short enough so that they compete.

erikh:
This rearchitects the volsupervisor and many things around it.

A limited changelog:

* Add functionality to take multiple locks at once
  * lock.ExecuteWithMultiUseLock() takes multiple use locks, a wait attribute,
    and a timeout.
  * Snapshot locks are now in use:
    * remove operations now take a snapshot lock as well as a mount lock; this
      allows it to remove volumes and their snapshots.
* Watch interface changes:
  * watch.Watch is now the core unit of transfer; each function is required to
    set Config on its way out.
* Config changes:
  * config.WatchGlobal uses watch.Watch interface
  * config.WatchVolume is new and uses the watch.Watch interface. Consumed by
    volsupervisor.
* Volsupervisor:
  * Rewritten from the ground up
  * Now schedules each snapshot create/prune in a goroutine, allowing multiple,
    independent snapshots to be taken at the same time.

Fixes #156


erikh:
fixes #156

unclejack:
LGTM

erikh:
@mapuri is busy today so I'm just going to merge this so we can move forward.

PR #179, opened 2016-03-14
Use exec

erikh:
requires #178 #176

This implements functionality to execute a command under use lock.


volcli/volcli.go
```diff
@@ -567,3 +572,73 @@ func useTheForce(ctx *cli.Context) (bool, error) {
...
+
+	args := ctx.Args()[1:]
+	if args[0] == "--" {
+		if len(args) < 2 {
+			return true, fmt.Errorf("You must supply a command to run")
+		}
+		args = args[1:]
+	}
+
+	err = lock.NewDriver(cfg).ExecuteWithMultiUseLock([]config.UseLocker{um, us}, true, 0, func(ld *lock.Driver, uls []config.UseLocker) error {
```
mapuri: just wondering, should be pick the timeout from `GlobalConfig` instead of fixing it to zero?

mapuri:
LGTM, this can pretty nice debug utility to test lock issues I think.

volcli/volcli.go
erikh: Not in this case. We want to hang as long as the user doesn't SIGINT us.

mapuri: ah, ok makes sense

Issue #181, opened 2016-03-16
Re-upload of policy orphans volumes


erikh:
reproduction:

```
volcli policy upload policy1 < /testdata/intent1.json
volcli volume create policy1/test
volcli policy upload policy1 < /testdata/fastsnap.json
volcli volume create policy1/test2
volcli volume list-all
<output>
policy1/test2
</output>
```

The resulting `policy1/test` is still in existence, but its etcd record is gone, basically orphaning it.

This should be an easy patch so I would ask that some others on the team could take it?

mapuri:
@erikh I can look into this. I am assuming the correct behavior would be to fail the second upload if policy with that name already exists, right?

mapuri:
ah, nvm I think I got it wrong. This is the case where user wants to update the intent so the volumes with old intent should have continued to exist.

erikh:
Correct on the last post; we re-upload the policy without trashing the volumes.

PR #183, opened 2016-03-16
systemtest for policy update

mapuri:
@erikh this is just the test for #181 

Somehow for me the test succeeds, so wanted to get your review once in case I am missing something wrt reproduction. In any case having this test seems like good.


erikh:
I looked into this, and I think this happened while I was implementing the snapshot copy features so I may have broken something there instead. I will keep an eye out for it, but let's resolve it for now with this test.

Merge when CI completes if you see it before I do.

fixes #181

erikh:
looks like something bugged out here, can you re-push?

mapuri:
yes, I pushed it again. I think this might have been related to my go version update on CI nodes.

PR #184, opened 2016-03-16
Volume snapshot copy

erikh:
This implements `volcli volume snapshot` commands:

* `volcli volume snapshot list` will list all snapshots for a given volume
* `volcli volume snapshot copy` will copy a snapshot from volume A into new volume B (under the same policy)

Please take a close look here as a lot has changed.

/cc @jainvipin @mapuri @unclejack

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -15,10 +15,11 @@ import (
 
 	"github.com/contiv/volplugin/executor"
 	"github.com/contiv/volplugin/storage"
+
+	log "github.com/Sirupsen/logrus"
 )
 
 const (
-	deviceBase = "/dev/rbd"
 	// BackendName is string for ceph storage backend
 	BackendName = "ceph"
 )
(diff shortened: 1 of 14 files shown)


erikh:
@mapuri this is ready for review when you have time.

storage/backend/ceph/ceph.go
```diff
@@ -321,6 +320,32 @@ func (c *Driver) ListSnapshots(do storage.DriverOptions) ([]string, error) {
...
+		return err
+	}
+
+	if er.ExitStatus != 0 {
+		return fmt.Errorf("Protecting snapshot for clone (volume %q, snapshot %q): %v", do.Volume.Name, snapName, err)
+	}
+
+	cmd = exec.Command("rbd", "clone", do.Volume.Name, newName, "--snap", snapName, "--pool", do.Volume.Params["pool"])
+	er, err = executor.NewWithTimeout(cmd, do.Timeout).Run()
+	if err != nil {
```
mapuri: I think we should unprotect the sanpshot if clone fails, right?

On a related note, does the snapshot needs to stay protected after clone is successful?

mapuri:
@erikh 

from ux and REST endpoint (`/copy`) perspective I think another alternative would have been to add a flag to the `volume create` sub-command itself, like:

```
volcli volume create --from-snap <> <policy>/<volume>
```

I wondering if the above UX is better as when we integrate this capability in the docker's storage-plugin API then we could just pass `--opt from-snap=<>` in `docker volume create`, which we could then pass to the volplugin's volume create call?

Also I don't see a docker side of integration in this PR, should we add it here in case it helps us make a more informed decision? Or do you plan to add it later?

storage/backend/ceph/ceph.go
erikh: correct on the unprotect.

yes, the snapshot cannot be unprotected once the copy is made

erikh:
That's not going to work. The reason is, we need to copy the volume to a new policy before, and the snap id itself is scoped to the existing volume. This won't work.

That said, we could extend it in the future to work this way with a combination of options (for source and snap id). I don't think it's necessary right now, though.

mapuri:
@erikh 

> The reason is, we need to copy the volume to a new policy before, and the snap id itself is scoped to the existing volume

may be i missed this, where does this happens? from my reading of `(d *DaemonConfig) handleCopy()` I was under the impression we create a new volume with exact same policy config as that of the volume to which the snapshot belongs.

so, just to clarify I got it correct, following work flow will work and create `vol2` with old version of policy:

```
volcli upload policy1 < intent-big-size-with-snap
volcli volume create policy1/vol1
volcli upload policy1 < intent-small-size
volcli volume snapshot copy policy1/vol1 snap1 vol2
```

> we could extend it in the future to work this way with a combination of options (for source and snap id). I don't think it's necessary right now, though.

sgtm

erikh:
Correct.

mapuri:
ok, thanks for confirming @erikh . We can revisit UX later when we integrate this feature with docker's API.

LGTM

PR #189, opened 2016-03-18
Vendor executor

erikh:
This vendors executor from github.com/contiv/executor and removes it from the tree.


mapuri:
LGTM

erikh:
I'm just goign to merge this sans CI; it's going to break because of the collins fix which comes after this PR. I know this PR is fine.

PR #192, opened 2016-03-18
Use errored

erikh:
This uses the new github.com/contiv/errored package. It will be implemented a little more hardy once I make some modifications to both ends, but this is a good enough approach for now.


mapuri:
LGTM

PR #197, opened 2016-03-18
Fix unmap

erikh:
This creates a number of fixes for potential and infrequent failures in the mount/unmount process in volplugin. It also exposes a bug in docker via https://github.com/docker/docker/pull/21015 which is required to fix the issue for multiple atttempts at unmounting the same mountpoint on the same host.

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -394,10 +394,6 @@ func (c *Driver) Mounted(timeout time.Duration) ([]*storage.Mount, error) {
 		return nil, err
 	}
 
-	if len(hostMounts) != len(mapped) {
-		return nil, errored.Errorf("Mounted and mapped volumes do not align.")
-	}
-
 	mounts := []*storage.Mount{}
 
 	for _, hostMount := range hostMounts {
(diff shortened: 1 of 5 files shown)


mapuri:
LGTM, I guess we need this as ceph won't allow mounting same volume more than once, right?

Just to understand the docker issue and the fix well, basically docker will continue to send unmount but has added an ID to mount and unmount request that volplugin can use to disregard unmount requests that don't have corresponding successful mount? Do we need to worry about these ID's being global v/s host local, looks like not becasue swarm will anyways create a volume on every host. Is that correct?

erikh:
It's just not safe to do, ceph will let it happen though. There is a facility in ceph itself to make/remove these locks, but it hasn't proven very reliable to me yet.

mapuri:
ah..ok, yeah I remember we had tried ceph locks before in volplugin.

erikh:
```
Just to understand the docker issue and the fix well, basically docker will continue to send unmount but has added an ID to mount and unmount request that volplugin can use to disregard unmount requests that don't have corresponding successful mount? Do we need to worry about these ID's being global v/s host local, looks like not becasue swarm will anyways create a volume on every host. Is that correct?
```

To answer this, yes, that is the exact scenario we are hitting.

erikh:
global vs. host local: these ids are generated locally and are per-mount, not per-volume, well at least once they arrive that is.

PR #198, opened 2016-03-20
More fixes

erikh:
This branch contains a much more emboldened battery test suite (50 containers x 3 nodes + 50 volumes) that exposed a number of races that only appear under stress. I have almost all of them shaked out now. 

There are a few small fixes in here too, as well as some code cleanup/simplification around the system tests and ceph driver specifically.

### DIFF
--- a/lock/client/client.go
+++ b/lock/client/client.go
@@ -107,12 +107,12 @@ func (d *Driver) reportMountEndpoint(endpoint string, ut *config.UseMount) error
 		return errored.Errorf("Could not read response body: %v", err)
 	}
 
-	if resp.StatusCode != 200 {
-		return errored.Errorf("Status was not 200: was %d: %q", resp.StatusCode, strings.TrimSpace(string(content)))
-	}
-
 	if resp.StatusCode == 404 {
 		return errNotFound
+	}
+
+	if resp.StatusCode != 200 {
+		return errored.Errorf("Status was not 200: was %d: %q", resp.StatusCode, strings.TrimSpace(string(content)))
 	}
 
 	return nil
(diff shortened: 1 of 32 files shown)


erikh:
I want to clean up this patch a little bit before review, I just wanted to see what happens when I send the new battery to CI.

lock/client/client.go
```diff
@@ -102,6 +111,10 @@ func (d *Driver) reportMountEndpoint(endpoint string, ut *config.UseMount) error
 		return errored.Errorf("Status was not 200: was %d: %q", resp.StatusCode, strings.TrimSpace(string(content)))
 	}
 
+	if resp.StatusCode == 404 {
```
mapuri: I think this code is unreachable as we would return earlier due to the check at line 110

systemtests/battery_test.go
```diff
@@ -28,150 +28,181 @@ func (s *systemtestSuite) TestBatteryMultiMountSameHost(c *C) {
 		}()
 	}
 
+	var realErr error
+
 	for x := 0; x < count; x++ {
-		c.Assert(<-errChan, NotNil)
+		err := <-errChan
+		if err != nil {
+			realErr = err
```
mapuri: I might be reading this wrong. But I think we should `break` on first `non-nil` error, otherwise `realErr` might get overwritten when a command that finishes last returns a nil error

mapuri:
LGTM, just a few minor comments.

lock/client/client.go
erikh: thanks.

systemtests/battery_test.go
erikh: that's ok; we need to fully exhaust the channel which is more important.

erikh:
@mapuri @unclejack PTAL ASAP. This branch has been maturing for a while and will need careful review.

mapuri:
LGTM

unclejack:
LGTM

PR #203, opened 2016-03-25
vendor the recent ansible repo

mapuri:
also bump up box version to work with new ansible

/cc @erikh


mapuri:
@erikh I just noticed that #202 also bring in new ansible. But may be we shall rebase that over this. This resolves the merge conflicts with subtree and also cleansup dependence of `volplugin-node` host-group on `dev` role that shall not be needed with new vagrant box. wdyt?

erikh:
I thikn I have this solved in the swarm pull request, but let me know if 
I don't.

-Erik

mapuri:
@erikh 

This PR brings in two changes on top. One is to revert volplugin-node host-group to not depend on `dev` role and moves volplugin's Vagrantfile to use the new box which works with this ansible.

I think if we rebase swarm PR on top of this, the respective ansible PR (https://github.com/contiv/ansible/pull/147) might also benefit wrt to the commit and merge history

erikh:
ok

PR #204, opened 2016-03-26
Runtime configuration

erikh:
This changes the json structure to be less verbose and more descriptive. It also partitions the parameters into several portions over the single `default-options` before, allowing for better handling of runtime configuration vs. create time configuration, as well as passing static driver parameters.

Please see the testdata for the new format.

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -11,16 +11,17 @@ import (
 // Policy is the configuration of the policy. It includes default
 // information for items such as pool and volume configuration.
 type Policy struct {
-	DefaultVolumeOptions VolumeOptions     `json:"default-options"`
-	FileSystems          map[string]string `json:"filesystems"`
+	CreateOptions  `json:"create"`
+	RuntimeOptions `json:"runtime"`
+	DriverOptions  map[string]string `json:"driver"`
+	FileSystems    map[string]string `json:"filesystems"`
+	Backend        string
 }
 
 // NewPolicy return policy config with specified backend preset
 func NewPolicy(backend string) *Policy {
 	return &Policy{
-		DefaultVolumeOptions: VolumeOptions{
-			Backend: backend,
-		},
+		Backend: backend,
 	}
 }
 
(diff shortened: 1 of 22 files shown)


erikh:
This needs to be rebased against #198 so it gets committed correctly.

erikh:
This is ready for review, @mapuri @unclejack

config/policy.go
```diff
@@ -11,16 +11,17 @@ import (
 // Policy is the configuration of the policy. It includes default
 // information for items such as pool and volume configuration.
 type Policy struct {
-	DefaultVolumeOptions VolumeOptions     `json:"default-options"`
-	FileSystems          map[string]string `json:"filesystems"`
+	CreateOptions  `json:"create"`
```
mapuri: `CreateOptions` seems to have `Filesystem` field, it is not relevant in Policy config, right?

mapuri:
@erikh the split config LGTM. It organizes config better than before.

I have a couple of design/behavior questions around the split that I thought I will clarify. I see that you have put create, runtime and driver options in both policy and volume. And I am wondering how do you see users would interact with these in following scenarios.
- Can a user still modify create time options of a policy as today?
- Would changing runtime config in a policy affect existing volumes? Or user will be now allowed to change it in the volume itself?

erikh:
To answer your questions:

1. No, not anymore without heading through the policy system. This decision was made because it was hard to clearly support situations where double-create happens, but I think now I have a better solution for this moving forward, I just have to implement it. Refusing double-create is something we couldn't do before, but can now. However, this needs to be done in incremental steps.
1. There are no runtime configs in policy, at least in no different way than there were before. They do not cascade and you will only be able to modify only the volume parameters in the newest set of patches in the PR pipeline.

config/policy.go
erikh: FileSystems was removed, sorry I wasn't clearer about that. It no longer makes sense with the create flags missing.

mapuri:
SGTM, thanks for clarifying @erikh

PR #209, opened 2016-04-04
Runtime configuration 4

erikh:
The biggest change here is that double-creates will fail if coming from volcli (docker will swallow the create for us in the `docker volume create` case). Additionally, I have re-integrated volume create flags, so that you can set up initial parameters for the filesystem, etc.

### DIFF
--- a/config/volume.go
+++ b/config/volume.go
@@ -18,23 +18,29 @@ import (
 // Volume is the configuration of the policy. It includes pool and
 // snapshot information.
 type Volume struct {
-	PolicyName string         `json:"policy"`
-	VolumeName string         `json:"name"`
-	Options    *VolumeOptions `json:"options"`
+	PolicyName     string            `json:"policy"`
+	VolumeName     string            `json:"name"`
+	DriverOptions  map[string]string `json:"driver"`
+	CreateOptions  `json:"create"`
+	RuntimeOptions `json:"runtime"`
+	Backend        string
 }
 
-// VolumeOptions comprises the optional paramters a volume can accept.
-type VolumeOptions struct {
-	Pool         string          `json:"pool" merge:"pool"`
-	Size         string          `json:"size" merge:"size"`
-	UseSnapshots bool            `json:"snapshots" merge:"snapshots"`
-	Snapshot     SnapshotConfig  `json:"snapshot"`
-	FileSystem   string          `json:"filesystem" merge:"filesystem"`
-	Ephemeral    bool            `json:"ephemeral,omitempty" merge:"ephemeral"`
-	RateLimit    RateLimitConfig `json:"rate-limit,omitempty"`
-	Backend      string          `json:"backend"`
+// CreateOptions are the set of options used by volmaster during the volume
+// create operation.
+type CreateOptions struct {
+	Size       string `json:"size"`
+	FileSystem string `json:"filesystem"`
 
 	actualSize units.Base2Bytes
+}
+
+// RuntimeOptions are the set of options used by volplugin when mounting the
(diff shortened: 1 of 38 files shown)


erikh:
This is the final branch, PTAL and review thoroughly @mapuri @unclejack

erikh:
missing a component here; this branch isn't finished yet.

erikh:
This branch is ready for review; please take a very close look. I have closed the other PRs so you only have to review this one.

@mapuri @unclejack

volplugin/handlers.go
```diff
@@ -313,13 +210,14 @@ func (dc *DaemonConfig) mount(w http.ResponseWriter, r *http.Request) {
...
 
-	if err := applyCGroupRateLimit(volConfig, mc); err != nil {
+	if err := applyCGroupRateLimit(volConfig.RuntimeOptions, mc); err != nil {
 		httpError(w, "Applying cgroups", err)
 		return
 	}
 
 	go dc.Client.HeartbeatMount(dc.Global.TTL, ut, dc.Client.AddStopChan(uc.Request.Name))
 
+	go dc.startRuntimePoll(volConfig.String(), mc)
```
mapuri: In addition to handling mount, in case of volplugin restart we may need to poll for already mounted volumes as well, right?

mapuri:
LGTM other than one clarification.

volplugin/handlers.go
erikh: OH god, yes.

erikh:
@mapuri this is ready for review again, I resolved the issue and found a large bug in how use locks work with this new system along the way. PTAL.

@unclejack PTAL as well

unclejack:
LGTM

config/volume.go
```diff
@@ -171,18 +214,20 @@ func (c *Client) ListVolumes(policy string) (map[string]*Volume, error) {
 	configs := map[string]*Volume{}
 
 	for _, node := range resp.Node.Nodes {
-		if node.Value == "" {
-			continue
-		}
+		if len(node.Nodes) > 0 {
+			node = node.Nodes[0]
+			if !node.Dir && strings.HasSuffix(node.Key, "/create") {
```
mapuri: if I am reading this correct, looks like ListVolumes() will now just show Create Config related info for a volume. Shouldn't we show both? I understand that we have a separate API to show runtime configuration, but I am wondering if we need a CLI or UX to show consolidated info as well.

config/volume.go
```diff
@@ -208,12 +253,48 @@ func (c *Client) ListAllVolumes() ([]string, error) {
...
+
+			w.Channel <- vw
+		}
+	})
+
+	watch.Create(w)
+}
+
+// WatchVolumeCreates watches the volumes tree and returns data back to the activity channel.
+func (c *Client) WatchVolumeCreates(activity chan *watch.Watch) {
```
mapuri: just curious, why do we need to watch create config? They can't change right?

mapuri:
@erikh , changes LGTM. I just a have a few clarifications.

config/volume.go
erikh: This is unused code I think, but it was just intended to provide completeness to the interface. Perhaps I should remove it?

mapuri: ah, ok..thanks for clarifying Erik.

Less code is better, few things that can break :). But if it a lot of effort to remove this, I am fine leaving it as is.

erikh:
Tests are running, but I've resolved the issues; PTAL

mapuri:
@erikh thanks for taking care of the comments.

LGTM

PR #212, opened 2016-04-08
config,systemtests,volplugin: handle a number of situations where our…

erikh:
… model doesn't match docker's:

* fixup the systemtests to enable the multi mount same host test, improve some code quality.
* Use locks now verify that the content exists before comparing it.
* Handle "volume not found" errors in a way that lets docker consume it.
* Resolve a number of situations where input was not error checked, just silently returned.
* Use a reference counter to refuse mounts in situations where docker attempts
  to unmounts devices that belong to other containers.

### DIFF
--- a/systemtests/battery_test.go
+++ b/systemtests/battery_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-	"time"
 
 	. "gopkg.in/check.v1"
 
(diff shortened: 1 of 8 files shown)


erikh:
@mapuri @unclejack another one I'd like to get in today, if you have time PTAL. Another big-ish one complexity-wise.

systemtests/battery_test.go
```diff
@@ -13,40 +12,70 @@ import (
...
+		for x := 0; x < count; x++ {
+			<-syncChan
+		}
 
-	for x := 0; x < count; x++ {
-		err := <-errChan
-		if err != nil {
-			realErr = err
+		// FIXME netplugin is broken
+		c.Assert(s.restartNetplugin(), IsNil)
```
mapuri: I might be missing something, but somehow I don't see any netplugin specific logic in this test. How does netplugin interfere with volpluglin here?

mapuri:
@erikh LGTM

For my understanding, the new `mountCount` logic is to handle duplicate mounts coming locally from docker to the volplugin, right? i.e. mount for same volume across hosts will still be prevented by the use.Lock functionality?

erikh:
Correct.

systemtests/battery_test.go
erikh: netplugin panics occasionally for some reason when I run this test; I hacked around it for now.

erikh: /cc @shaleman

mapuri:
thanks @erikh 

one more follow up instead of new `mountCount` logic, why not just use `use.Lock` locally as well? I am thinking it will reuse existing functionality and keep logic simple but I am not sure if I am thinking it through.

erikh:
This doesn't work because the locks are per-volume, not per-mount (and to be exact, it would be nearly impossible to manage this with per-mount locking)

erikh:
I'm going to merge this, let me know if you'd like to see any corrections and I can put them in a separate PR.

PR #213, opened 2016-04-09
Fix ansible

erikh:
This moves ansible to use a submodule instead of subtree. given the state of flux in both repositories, this seems like the better idea.


erikh:
I'm merging this

PR #219, opened 2016-04-12
enable volcli systemtests with null driver

mapuri:
these changes:
- add some instrumentation to all different backend drivers  to be tested against same set of tests by passing driver name in the environment
- update the system-test target to run volcli tests for null driver
- update the null driver implementation to keep state in order to enable running tests against it

/cc @erikh PTAL when you get a chance. I will see if I can make other system tests work too but no promises atm as they might not make it to this PR :)


erikh:
gonna have to re-push, I broke one of the slaves earlier today it seems.

Makefile
```diff
@@ -103,7 +103,18 @@ run-build: godep
...
+		GOPATH=/tmp/volplugin:${WORKSPACE} godep go test -v -timeout 240m ./systemtests -check.v; \
+	else \
+		godep go test -v -timeout 240m ./systemtests -check.v; \
+	fi
+	@echo running null-driver tests...
+	if [ "x${WORKSPACE}" != "x" ]; then \
+		USE_DRIVER="null" GOPATH=/tmp/volplugin:${WORKSPACE} godep go test -v -timeout 240m ./systemtests -check.v -check.f systemtestSuite.TestVolCLI*; \
+	else \
+		USE_DRIVER="null" godep go test -v -timeout 240m ./systemtests -check.v -check.f systemtestSuite.TestVolCLI*; \
+	fi
```
erikh: can we move this into a shell script? maybe put it under `build`?

systemtests/util_test.go
```diff
@@ -158,9 +163,25 @@ func (s *systemtestSuite) rebootstrap() error {
...
+}
+
+func getDriverDir() string {
+	driverDir := ""
+	if getDriver() != "ceph" {
+		driverDir = getDriver()
+	}
+	return driverDir
+}
+
```
erikh: we should not do this; we should provde the `testdata` dirs for ceph too, and avoid this guard and extra code.

mapuri: yes, makes sense. Let me make the change.

Makefile
mapuri: sure, good idea let me take this out into script.

mapuri:
@erikh I will push a new diff with comments included in a bit.

Also I wanted your confirmation about this line (https://github.com/contiv/volplugin/pull/219/files#diff-e39a52a55ff9b6e527b965ae02b0264aR219). Is it ok to test the duplicate create failure only for ceph? Or do you think all drivers shall behave like ceph?

erikh:
I think it's ok.

mapuri:
@erikh this is ready for another pass. PTAL when you get a chance

storage/backend/null/null.go
```diff
@@ -1,104 +1,162 @@
...
 func (d *Driver) List(lo storage.ListOptions) ([]storage.Volume, error) {
+	log.Infof("In %q", getFunctionName())
 	return []storage.Volume{}, nil
 }
 
-// Mount is a noop
+// Mount records mounted volume
 func (d *Driver) Mount(do storage.DriverOptions) (*storage.Mount, error) {
+	log.Infof("In %q", getFunctionName())
+	if err := os.MkdirAll(d.mountpath, 0700); err != nil && !os.IsExist(err) {
```
erikh: I know this is a lot more work, but I really think this shouldn't write to the filesystem. Perhaps we should instead map the paths and check for path existence instead, and return a useful error. This way no disk writes will need to occur. Also, I notice these paths are not getting cleaned up in Unmount.

erikh:
Ok this LGTM. If you want, wrap `getDriver()` in the systemtests with `ceph()` which performs the comparison against string.

Then you can just:

```
if ceph() {
}
```

storage/backend/null/null.go
mapuri: > I really think this shouldn't write to the filesystem. Perhaps we should instead map the paths and check for path existence instead, and return a useful error. 

@erikh just to understand, do you mean we should not create the `d.mountpath` and `volumePath` directories at all? OR just create them once and avoid subsequent `os.MkdirAll()` calls.

IIRC we do need to create these directories otherwise `docker run` complains. I think I can optimize by checking for existence of the directory (by keeping local state) and not call `os.MkdirAll()` at all. Does that work?

> Also, I notice these paths are not getting cleaned up in Unmount.

yeah, that was part intentional and part laziness. Basically by not having an explicit `rmdir` call on every unmount I saved some work (both mine and system's :) ) of not making the extra system call. But thinking more adding the extra call might be better as it cleans up state after a successful test. Or we can do it just once like we do in `purgeVolumes()` for ceph. Latter might be better, wdyt?

mapuri:
@erikh 

> wrap getDriver() in the systemtests with ceph() which performs the comparison against string.

This is a good suggestion. I will make this change in next push.

storage/backend/null/null.go
erikh: Oh, I didn't even consider docker's needs, LOL.

Let's just remove the paths at Unmount time and call it a day.

systemtests/util_test.go
```diff
@@ -158,9 +163,21 @@ func (s *systemtestSuite) rebootstrap() error {
 	return nil
 }
 
+func getDriver() string {
+	driver := "ceph"
+	if strings.Trim(os.Getenv("USE_DRIVER"), " ") != "" {
```
erikh: ok, sorry to throw this back at you AGAIN. Why are we not using `strings.TrimSpace` here?

erikh:
LGTM other than my question/request

systemtests/util_test.go
mapuri: I missed the built in TrimSpace() function. Let me use that here instead.

mapuri:
@erikh all comments should be incorporated now. PTAL

erikh:
LGTM

PR #220, opened 2016-04-14
Restructure storage drivers

erikh:
This provides an implementation of a stub NFS driver + a revisitation of storage drivers in general.

Drivers are now split into Mount, Snapshot, and CRUD (Create/Retrieve/Update/Delete) drivers. Each driver has its own unique sets of methods (except for Name, which is shared through `storage.NamedDriver`) that do not overlap. To operate with a driver, you use the `backend` library to take it, then perform operations on it based on the standard interfaces.

Now, drivers do not have to provide functionality for every subsystem, just the ones they support. Future patches will implement this integration as we get more storage capabilities.

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -2,7 +2,6 @@ package ceph
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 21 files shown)


storage/backend/ceph/ceph.go
```diff
@@ -46,52 +45,70 @@ func runWithTimeout(cmd *exec.Cmd, timeout time.Duration) (*executor.ExecResult,
...
 		if er.ExitStatus == 17 {
 			return storage.ErrVolumeExist
 		} else if er.ExitStatus != 0 {
-			return errored.Errorf("Creating disk %q: %v", do.Volume.Name, er)
+			return errored.Errorf("Creating disk %q: %v", intName, er)
 		}
 	} else if err != nil {
-		fmt.Println(err.(*exec.ExitError).ProcessState.Sys().(syscall.WaitStatus))
-		return errored.Errorf("%#v", err)
+		return errored.Errorf("Cerating Disk: %#v", err)
```
mapuri: s / Cerating / Creating

mapuri:
LGTM

The new structure provides a good way to implement drivers with different capabilities.

erikh:
rebased against your work, PTAL at the null implementation specifically.

mapuri:
LGTM

erikh:
one more peek please -- I made some heavier refactors as I was merging things. This branch should also pass. :)

PR #221, opened 2016-04-14
build,storage,systemtests: rename 'null' driver to 'test' driver

erikh:


erikh:
goign to merge this as it's a pretty safe change.

PR #222, opened 2016-04-15
removing the stale --debug flag

vvb:


erikh:
this should probably be committed directly to ansible; we're going to replace  this tree soon with fixes.

vvb:
ok, will close this and raise a PR on the ansible tree.

PR #224, opened 2016-04-15
Null driver

erikh:
This implements a null (that is, no operations at all) driver suitable for use in production. it will be the foundation of the lack of a CRUD or snapshot system for NFS operations.

This also makes the backend dependent now on the volume instead of the global, which was removed. This is the next step towards multiple backends.

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -5,6 +5,7 @@ import (
 	"path"
 
 	"github.com/contiv/errored"
+	"github.com/contiv/volplugin/storage/backend/ceph"
 	"github.com/coreos/etcd/client"
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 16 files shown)


config/policy.go
```diff
@@ -20,9 +21,9 @@ type Policy struct {
 }
 
 // NewPolicy return policy config with specified backend preset
-func NewPolicy(backend string) *Policy {
+func NewPolicy() *Policy {
 	return &Policy{
-		Backend: backend,
+		Backend: ceph.BackendName,
```
mapuri: shouldn't we remove the `Backend` field from policy altogther?

mapuri: My previous comment might not be very clear. I am wondering if we shall just accept backend at time of `volume create` and default to `ceph` just there?

erikh: No. The volume needs a jumping-off-point to create the volume from, and 
that Backend is really it, so policy determines backend, volume just 
carries it.

mapuri:
got, it thanks @erikh 

LGTM

PR #227, opened 2016-04-15
Vagrantfile,volplugin: update to docker 1.11, remove some hacks needed in 1.10

erikh:


mapuri:
LGTM

@erikh so now volplugin will only work with docker 1.11+? I think it should be fine but just confirming.

erikh:
Yeah.

PR #231, opened 2016-04-18
*: split policy drivers for end-user

erikh:

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -13,17 +13,25 @@ import (
 // Policy is the configuration of the policy. It includes default
 // information for items such as pool and volume configuration.
 type Policy struct {
+	Name           string            `json:"name"`
 	CreateOptions  CreateOptions     `json:"create"`
 	RuntimeOptions RuntimeOptions    `json:"runtime"`
 	DriverOptions  map[string]string `json:"driver"`
 	FileSystems    map[string]string `json:"filesystems"`
-	Backend        string            `json:"backend"`
+	Backends       BackendDrivers    `json:"backends"`
+}
+
+// BackendDrivers is a struct containing all the drivers used under this policy
+type BackendDrivers struct {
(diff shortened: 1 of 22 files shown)


erikh:
@mapuri can you review this when you're available?

mapuri:
@erikh 

changes LGTM, exposing the volplugin capabilities all the way to user gives them more control which is really flexible.

A few comments/questions on new behavior:
- this flexibility allows user to mix and match incompatible drivers. I am assuming we will let backend implementations fail appropriately. For instance, say CRUD is specified as ceph and Mount is nfs, it is not expected to work, right? It's not an issue as long as users know the compatible backends but I am more worried about catching and testing this in backends. The combinations are huge. May be we can fail early at the API invocation itself by validating that user only chooses compatible drivers?
- On similar lines, if user specifies a policy with just `Mount` driver mentioned as `ceph` then they won't be able to use this policy for snapshots without deleting and recreating it. Again not a big issue but something that user will need to be aware of. Should we instead tie drivers/backends with capabilities and initialize the `backendsDrivers` structure based on capabilities of the backend that user specifies? This won't allow mixing different backends though but I am wondering if we need that?

erikh:
Yeah, I was thinking something similar when I started reading your reply :) Establish a compatibility matrix through driver acceptance (e.g., a map of driver -> supported pairs). This will be a pretty complex patch, though.

erikh:
To be clear, that should be addressed in a separate patch, perhaps after the NFS feature is complete. Do you mind putting in an issue with your thoughts on the topic?

mapuri:
> Establish a compatibility matrix through driver acceptance (e.g., a map of driver -> supported pairs

yeah, I think this keep rest of functionality simple and puts the complex compatibility check just at API handling level.

> Do you mind putting in an issue with your thoughts on the topic?

sgtm. Let me file an issue.

PR #236, opened 2016-04-18
systemtests,volcli,config,volsupervisor: perform snapshots based on a file written to etcd.

### LINKED ISSUE #218: Volsupervisor to take a signal to snapshot remotely
volsupervisor needs a HTTP REST endpoint to service a signal from applications or database containers which wish to sync their filesystems before snapshotting for a consistent snapshot; this signal would create a manual snapshot that would still get cleaned up in a normal `keep` scenario.

The signaling will be toggle-able through the policy by adding a `signal: true` property in the snapshot runtime configuration. It is orthogonal to the standard snaps (which should now accept a `frequency` of null to turn them off, but still allow for `keep` pruning)

It would also be nice if we had a tool that accomplished the client side of this operation, e.g., "run a command then fire the snapshot, then run another command". Wrapping this will play well into the security story later.

erikh:
fixes #218

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -10,12 +10,13 @@ import (
 )
 
 const (
-	rootVolume = "volumes"
-	rootUse    = "users"
-	rootPolicy = "policies"
+	rootVolume    = "volumes"
+	rootUse       = "users"
+	rootPolicy    = "policies"
+	rootSnapshots = "snapshots"
 )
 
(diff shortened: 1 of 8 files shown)


mapuri:
LGTM

do user created snapshots need any special handling wrt `pruneSnapshot` logic? Right now I think they will be handled at par with the `snapshot.Keep` policy, right?

erikh:
@mapuri yeah

PR #255, opened 2016-05-10
volmaster,volplugin: make routes more consistent

### LINKED ISSUE #244: REST endpoint to list all the policies
It will be nice if it returns an array of policies:
```
[
{
"name": "db-policy",
 "create": {
   "size": "10MB",
   "filesystem": ""
 },
 "runtime": {
   "snapshots": true,
   "snapshot": {
     "frequency": "30m",
     "keep": 20
   },
   "rate-limit": {
     "write-iops": 0,
     "read-iops": 0,
     "write-bps": 0,
     "read-bps": 0
   }
 },
 "driver": {
   "pool": "rbd"
 },
 "filesystems": {
   "ext4": "mkfs.ext4 -m0 %"
 },
 "backend": "test"
}
]
```

unclejack:
Closes #244

### DIFF
--- a/volcli/commands.go
+++ b/volcli/commands.go
@@ -2,15 +2,6 @@ package volcli
 
 import "github.com/codegangsta/cli"
 
-// VolmasterFlags contains the flags specific to volmasters
-var VolmasterFlags = []cli.Flag{
-	cli.StringFlag{
-		Name:  "volmaster",
-		Usage: "address of volmaster process",
-		Value: "127.0.0.1:9005",
-	},
-}
-
 // GlobalFlags are required global flags for the operation of volcli.
 var GlobalFlags = []cli.Flag{
 	cli.StringFlag{
@@ -22,6 +13,11 @@ var GlobalFlags = []cli.Flag{
 		Name:  "etcd",
 		Usage: "URL for etcd",
 		Value: &cli.StringSlice{"http://localhost:2379"},
+	},
+	cli.StringFlag{
+		Name:  "volmaster",
+		Usage: "address of volmaster process",
+		Value: "127.0.0.1:9005",
 	},
 }
 
@@ -41,7 +37,6 @@ var Commands = []cli.Command{
 			},
 			{
 				Name:        "get",
-				Flags:       VolmasterFlags,
 				ArgsUsage:   "[policy name]",
 				Usage:       "Obtain the global configuration",
 				Description: "Gets the global configuration from etcd",
@@ -55,7 +50,6 @@ var Commands = []cli.Command{
 		Subcommands: []cli.Command{
 			{
 				Name:        "upload",
-				Flags:       VolmasterFlags,
 				ArgsUsage:   "[policy name]. accepts from stdin",
 				Description: "Uploads a policy to etcd. Accepts JSON. Requires direct, unauthenticated access to etcd.",
 				Usage:       "Upload a policy to etcd",
@@ -70,7 +64,6 @@ var Commands = []cli.Command{
 			},
 			{
 				Name:        "get",
-				Flags:       VolmasterFlags,
 				ArgsUsage:   "[policy name]",
 				Usage:       "Obtain the policy",
 				Description: "Gets the policy for a policy from etcd.",
@@ -91,10 +84,10 @@ var Commands = []cli.Command{
 		Subcommands: []cli.Command{
 			{
 				Name: "create",
-				Flags: append(VolmasterFlags, cli.StringSliceFlag{
+				Flags: []cli.Flag{cli.StringSliceFlag{
 					Name:  "opt",
 					Usage: "Provide key=value options to create the volume",
-				}),
+				}},
 				ArgsUsage:   "[policy name]/[volume name]",
 				Description: "This creates a logical volume. Calls out to the volmaster and sets the policy based on the policy name provided.",
 				Usage:       "Create a volume for a given policy",
@@ -133,7 +126,6 @@ var Commands = []cli.Command{
 				ArgsUsage:   "[policy name]/[volume name]",
 				Description: "Remove the volume for a policy, deleting its contents.",
 				Usage:       "Remove a volume and its contents",
-				Flags:       VolmasterFlags,
 				Action:      VolumeRemove,
 			},
 			{
@@ -144,7 +136,6 @@ var Commands = []cli.Command{
 					{
 						Name:        "take",
 						ArgsUsage:   "[policy name]/[volume name]",
-						Flags:       VolmasterFlags,
 						Description: "Take a snapshot for a volume now",
 						Usage:       "Take a snapshot for a volume now",
 						Action:      VolumeSnapshotTake,
@@ -154,7 +145,6 @@ var Commands = []cli.Command{
 						ArgsUsage:   "[policy name]/[volume name]",
 						Description: "List snapshots",
 						Usage:       "List snapshots",
-						Flags:       VolmasterFlags,
 						Action:      VolumeSnapshotList,
 					},
 					{
@@ -162,7 +152,6 @@ var Commands = []cli.Command{
 						ArgsUsage:   "[policy name]/[volume name] [snapshot name] [new volume name]",
 						Description: "Copies a volume with a given snapshot name to the new volume name. The policy will remain the same, as well as the volume parameters.",
 						Usage:       "Copy a volume snapshot to a new volume",
-						Flags:       VolmasterFlags,
 						Action:      VolumeSnapshotCopy,
 					},
 				},
@@ -177,7 +166,6 @@ var Commands = []cli.Command{
 						ArgsUsage:   "[policy name]/[volume name]",
 						Description: "Get runtime configuration",
 						Usage:       "Get runtime configuration",
-						Flags:       VolmasterFlags,
 						Action:      VolumeRuntimeGet,
 					},
 					{
--- a/volcli/volcli.go
+++ b/volcli/volcli.go
@@ -204,7 +204,7 @@ func policyGet(ctx *cli.Context) (bool, error) {
 
 	policy := ctx.Args()[0]
 
-	resp, err := http.Get(fmt.Sprintf("http://%s/policy/%s", ctx.String("volmaster"), policy))
+	resp, err := http.Get(fmt.Sprintf("http://%s/policies/%s", ctx.String("volmaster"), policy))
 	if err != nil {
 		return false, err
 	}
--- a/volcli/volcli.go
+++ b/volcli/volcli.go
@@ -4,12 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
-	"path"
 	"strings"
 	"syscall"
 
@@ -62,7 +62,7 @@ func GlobalGet(ctx *cli.Context) {
 }
 
 func queryGlobalConfig(ctx *cli.Context) (*config.Global, error) {
-	resp, err := http.Get(fmt.Sprintf("http://%s/global", ctx.String("volmaster")))
+	resp, err := http.Get(fmt.Sprintf("http://%s/global", ctx.GlobalString("volmaster")))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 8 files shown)


volmaster/daemon.go
```diff
@@ -83,8 +83,8 @@ func (d *DaemonConfig) Daemon(debug bool, listen string) {
 
 	getRouter := map[string]func(http.ResponseWriter, *http.Request){
 		"/policy/{policy}":             d.handlePolicy,
-		"/list":                        d.handleList,
-		"/get/{policy}/{volume}":       d.handleGet,
+		"/volume/list":                 d.handleList,
```
mapuri: should we use `volumes` instead of `volume`, just to be consistent with `snapshots` below?

mapuri:
@unclejack LGTM other than one minor comment.

volmaster/daemon.go
mapuri: oh, if we do this then same will apply for `policy` too.

erikh:
This is a good start, but not exactly what I was hoping for. That's not your fault; I have poorly specced it.

I will add more specifications to the original issue. This is a much larger change than I had projected, I suppose.

volmaster/daemon.go
```diff
@@ -70,25 +70,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
 	for path, f := range postRouter {
 		r.HandleFunc(path, logHandler(path, d.Global.Debug, f)).Methods("POST")
 	}
 
 	getRouter := map[string]func(http.ResponseWriter, *http.Request){
-		"/policy/{policy}":             d.handlePolicy,
-		"/list":                        d.handleList,
-		"/get/{policy}/{volume}":       d.handleGet,
+		"/config/global":               d.handleGlobal,
+		"/policies/{policy}":           d.handlePolicy,
```
erikh: what if I have a policy named `list`?

volmaster/daemon.go
```diff
@@ -70,25 +70,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
+		"/policies/list":               d.handlePolicyList,
+		"/volumes/listall":             d.handleListAll,
+		"/volumes/list/{policy}":       d.handleList,
+		"/volumes/{policy}/{volume}":   d.handleGet,
```
erikh: very similar here.

volmaster/daemon.go
```diff
@@ -70,25 +70,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
 
 	postRouter := map[string]func(http.ResponseWriter, *http.Request){
-		"/request":      d.handleRequest,
-		"/create":       d.handleCreate,
-		"/copy":         d.handleCopy,
-		"/mount":        d.handleMount,
-		"/mount-report": d.handleMountReport,
-		"/unmount":      d.handleUnmount,
-		"/remove":       d.handleRemove,
+		"/config/global":                    d.handleGlobalUpload,
```
erikh: We don't need a namespace for this as there's only one. Let's avoid adding anything to this route's text.

erikh:
Needs tests. The issues above alone need tests for them, e..g, we should verify that `volcli volume list list/myvolume` works, `listall/myvolume` and `volcli policy get list` should all work.

But it needs tests testing all the new REST functionality.

erikh:
is this ready for review?

erikh:
Conflicts. Rebase and I will review tonight.

unclejack:
@erikh It's done. PTAL.

erikh:
This LGTM. Nice work.

@mapuri @dseevr

volcli/commands.go
```diff
@@ -34,6 +34,7 @@ var Commands = []cli.Command{
 		Subcommands: []cli.Command{
 			{
 				Name:        "upload",
+				Flags:       VolmasterFlags,
```
mapuri: @unclejack I think `VolmasterFlags` is now better put in `volcli.GlobalFlags`, that way we don't have to repeat it for every sub-command. wdyt?

volcli/volcli.go
```diff
@@ -120,13 +120,14 @@ func globalUpload(ctx *cli.Context) (bool, error) {
 		return false, err
 	}
 
-	cfg, err := config.NewClient(ctx.GlobalString("prefix"), ctx.GlobalStringSlice("etcd"))
+	resp, err := http.Post(fmt.Sprintf("http://%s/global", ctx.String("volmaster")), "application/json", bytes.NewBuffer(content))
```
mapuri: minor: we should perhaps use a function to form the request URL (i.e. wrap `fmt.Sprintf` there), that way if we later change something (like url scheme from http to https, or volmaster name to apiserver or similar) we won't need to again touch all these places.

volcli/volcli.go
```diff
@@ -350,15 +377,31 @@ func volumeForceRemove(ctx *cli.Context) (bool, error) {
...
-	if err := cfg.RemoveVolume(policy, volume); err != nil {
+	resp, err := http.Post(fmt.Sprintf("http://%s/volumes/removeforce", ctx.String("volmaster")), "application/json", bytes.NewBuffer(content))
+	if err != nil {
 		return false, err
 	}
 
+	qualifiedVolume := strings.Join([]string{policy, volume}, "/")
+
+	if resp.StatusCode == 404 {
+		return false, errored.Errorf("Volume %v no longer exists.", qualifiedVolume)
```
mapuri: nit pick: should the error be `Volume %v doesn't exists.` or similar ?

volcli/volcli.go
```diff
@@ -350,15 +377,31 @@ func volumeForceRemove(ctx *cli.Context) (bool, error) {
...
+	}
+
+	if resp.StatusCode != 200 {
+		return false, errored.Errorf("Volume %v Response Status Code was %d, not 200: %v", qualifiedVolume, resp.StatusCode, resp.Status)
```
mapuri: should we dump the resp body on stderr here? I think this is true for other places as well. I see we dump the body only in few cases, but I feel we should dump it always in case of non 200 (or 404) return, usually the body will have more informative cues on the failure.

volmaster/daemon.go
```diff
@@ -71,25 +71,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
 	postRouter := map[string]func(http.ResponseWriter, *http.Request){
-		"/request":      d.handleRequest,
-		"/create":       d.handleCreate,
-		"/copy":         d.handleCopy,
-		"/mount":        d.handleMount,
-		"/mount-report": d.handleMountReport,
-		"/unmount":      d.handleUnmount,
-		"/remove":       d.handleRemove,
+		"/global":                           d.handleGlobalUpload,
+		"/volumes/create":                   d.handleCreate,
```
mapuri: I might be missing something, but looks like our `volumes` REST endpoints don't take the volume-name in the URL, but `policies` and `snapshots` endpoints take the policy or volume name in URL. Is this intentional?

volmaster/daemon.go
```diff
@@ -71,25 +71,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
+		"/volumes/copy":                     d.handleCopy,
+		"/volumes/remove":                   d.handleRemove,
+		"/volumes/removeforce":              d.handleRemoveForce,
+		"/volumes/request":                  d.handleRequest,
+		"/mount":                            d.handleMount,
```
mapuri: should `mount` and `unmount` be part of `volumes` hierarchy? i.e. `volumes/mount/` and `volumes/unmount/`

volmaster/daemon.go
```diff
@@ -71,25 +71,33 @@ func (d *DaemonConfig) Daemon(listen string) {
...
+		"/mount-report":                     d.handleMountReport,
+		"/policies/{policy}":                d.handlePolicyUpload,
+		"/policies/delete/{policy}":         d.handlePolicyDelete,
+		"/runtime/{policy}/{volume}":        d.handleRuntimeUpload,
```
mapuri: similar to mount/unmount may be `/runtime/{policy}/{volume}` should be `/volumes/runtime/{policy}/{volume}` ?

volmaster/daemon.go
erikh: I asked him not to do it. This is the locking code that travels through the volmaster and I didn't want it fiddled with.

volmaster/daemon.go
erikh: well, the volume endpoints can suss out the information from the JSON post. Do you feel it's necessary for another reason?

volmaster/daemon.go
mapuri: ah, ok. Got it

volcli/volcli.go
erikh: "no longer exists" is the correct english.

volcli/volcli.go
erikh: +1

volcli/commands.go
erikh: actually +1 to this too. It can be provided for every command so it's no longer necessary.

volcli/volcli.go
erikh: Yes, this will come in a later patch.

erikh: What I actually want is a URI object that we can munge more safely.

volmaster/daemon.go
mapuri: ah, I see, this should be fine as well.

I was just wondering wrt consistency of the endpoints. Right now, in policy and snapshot endpoints we are passing the resource name in the URL but for volumes we don't, so thought I will check.

volmaster/daemon.go
erikh: This was also requested to not be done because it's related to some 
critical etcd operations. The CRUD stuff is what I asked him to complete 
in this patch.

volmaster/daemon.go
erikh: Sure. So, I'm not too concerned about that right now because after 0.1 
we're going to gut this thing and replace it with something testable. :)

Let's make sure the code works and is tested, but for now at least, 
let's think/talk about what the future is like and maintain otherwise.

mapuri: yes, sgtm ! 

I also think until 1.0 we can change APIs in a backward incompatible way, so we will have some leeway to rethink some of this.

volmaster/daemon.go
mapuri: got it, sounds good.

volmaster/daemon.go
erikh: zactly

volcli/volcli.go
```diff
@@ -120,13 +120,14 @@ func globalUpload(ctx *cli.Context) (bool, error) {
...
-	cfg, err := config.NewClient(ctx.GlobalString("prefix"), ctx.GlobalStringSlice("etcd"))
+	resp, err := http.Post(fmt.Sprintf("http://%s/global", ctx.String("volmaster")), "application/json", bytes.NewBuffer(content))
 	if err != nil {
 		return false, err
 	}
 
-	if err := cfg.PublishGlobal(global); err != nil {
-		return false, err
+	if resp.StatusCode != 200 {
+		io.Copy(os.Stderr, resp.Body)
```
dseevr: Technically, this function can fail and we should be checking the error to make sure it's nil (not EOF).  There's four other spots it's called in this commit, as well.

dseevr:
LGTM aside from the noted things 👍

volcli/volcli.go
erikh: I don't really see this being a problem as our responses are so small, but we probably should just check & log and move on.

volcli/commands.go
unclejack: I've moved this to the GlobalFlags, but we'll still be forced to specify GlobalFlags for every sub-command. This change will be in the next code push.

volcli/commands.go
```diff
@@ -34,14 +30,15 @@ var Commands = []cli.Command{
 		Subcommands: []cli.Command{
 			{
 				Name:        "upload",
+				Flags:       GlobalFlags,
```
mapuri: @unclejack , this should not be needed I think as user can just do `volcli --volmaster=<> upload < ...`

mapuri:
LGTM, except for one clarification on specifying GlobalFlags with every sub-command. I think we shouldn't need it per sub-command.

volcli/commands.go
erikh: agreed. should just be ripping a few lines out, so this is worth doing now.

volcli/commands.go
```diff
@@ -2,15 +2,6 @@ package volcli
...
-	cli.StringFlag{
-		Name:  "volmaster",
-		Usage: "address of volmaster process",
-		Value: "127.0.0.1:9005",
-	},
-}
-
 // GlobalFlags are required global flags for the operation of volcli.
 var GlobalFlags = []cli.Flag{
 	cli.StringFlag{
```
erikh: we should also remove these lines if we no longer need to contact etcd.

erikh: /cc @unclejack @mapuri

unclejack: We still need that for the other commands which still use etcd.

erikh: gotcha

erikh:
@unclejack are we all good here on comments?

mapuri:
LGTM

PR #256, opened 2016-05-14
NFS driver

erikh:
This is the first implementation of the NFS driver, with the exclusive locking support. We will make locking optional in a second pass.

There are a lot of commits here and most of it is code cleanup and reorg, NOT the NFS driver. Please review carefully as a lot of subtleties are being adjusted (and hopefully, are commented better now).


jainvipin:
@erikh  👍

erikh:
If you guys could review this as soon as possible, I would appreciate it. Master is breaking right now because it needs the openstack fix.

erikh:
@mapuri @unclejack @dseevr ^

vagrant_variables.yml
```diff
@@ -17,4 +17,4 @@ memory: 8192
 #   - https://atlas.hashicorp.com/boxes/search?utf8=✓&sort=&provider=virtualbox&q=
 #   - https://download.gluster.org/pub/gluster/purpleidea/vagrant/
 vagrant_box: contiv/centos72
-box_version: 0.3.2
+box_version: 0.6.0
```
unclejack: Do we have both VirtualBox and VMware boxes for this version?

erikh:
@mapuri can you `host nowhere` on the CI boxes and tell me if it returns NXDOMAIN or not? I'm having an issue with one of the negative tests for NFS.

vagrant_variables.yml
erikh: Yep

storage/backend/null/null.go
```diff
@@ -2,10 +2,20 @@
```
unclejack: Do we need these changes on this branch? The previous commit was disabling these.

storage/backend/ceph/internals.go
```diff
@@ -156,7 +156,7 @@ retry:
 
 	if err := json.Unmarshal([]byte(er.Stdout), &rbdmap); err != nil {
 		log.Errorf("Could not parse RBD showmapped output, retrying: %s", er.Stderr)
-		time.Sleep(100 * time.Second)
+		time.Sleep(100 * time.Millisecond)
```
unclejack: This is great!

storage/backend/null/null.go
erikh: Yes, we need to incorporate them because they need to be finished, and @mapuri is already committed to doing this work.

storage/backend/ceph/internals.go
erikh: yes; @dseevr found it actually.

unclejack:
LGTM

vendor/github.com/vishvananda/netlink/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+before_script:
+  # make sure we keep path in tact when we sudo
```
dseevr: Typo here: "in tact" -> "intact"

dseevr: nvm just noticed this is under /vendor lol

dseevr:
Looks great 👍

erikh:
Presuming @mapuri gives me his approval on Monday, I will merge this then.

I also need to write docs, unless someone wants to volunteer for that.

mapuri:
looks good @erikh. 

I see that you had to disable test/null driver due to some issues, let's talk about those whenever you get some time so I could understand a bit more about them and see how you would like to bring them back.

Regarding `host nowhere` command on CI machines, it gives me command not found error. Do I need to install any package? package with name `host` package doesn't seem to exist.

erikh:
@mapuri I worked around that by using a non-existent IP, so no need. However, `bind-utils` is the package for reference.

Let's defnitely talk about the null driver.

Ok, merging this. Thanks everyone for the assistance.

🎉

PR #268, opened 2016-05-19
volmaster: correct an issue where a 500 would be sent in the event a 404 should be sent.

### LINKED ISSUE #263: volcli volume remove fails to remove rbd volume (500 error) but removes volume in etcd
This should not happen, perhaps the order should be reversed. `force-remove` solves the orphaned volplugin volume problem.

erikh:

fixes #263

### DIFF
--- a/volmaster/daemon.go
+++ b/volmaster/daemon.go
@@ -423,6 +423,22 @@ func (d *DaemonConfig) handleRemove(w http.ResponseWriter, r *http.Request) {
 		Reason: lock.ReasonRemove,
 	}
 
+	etcdRemove := func() error {
+		if err := d.Config.RemoveVolume(req.Policy, req.Volume); err != nil {
+			return errored.Errorf("Clearing volume records for %q", vc).Combine(err)
+		}
+
+		return nil
+	}
+
+	complete := func() error {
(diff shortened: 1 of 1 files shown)


volmaster/daemon.go
```diff
@@ -423,30 +423,38 @@ func (d *DaemonConfig) handleRemove(w http.ResponseWriter, r *http.Request) {
...
 			return err
 		}
 
 		if err == errNoActionTaken {
-			goto publish
+			return complete()
 		}
 
 		if !exists {
+			etcdRemove()
```
mapuri: I might be reading this wrong, but may be this should be `return etcdRemove()` ? Otherwise I think we will fall through and call `complete()` below which will call `etcdRemove()` again and fail.

erikh: no, it returns `config.ErrNotExist` which is the desired behavior, and ensures the record was removed from etcd. This is the actual problem here that is being fixed.

I should perhaps comment this logic. I don't like it, but it seems to want to fit in this way.

mapuri: ah, yes you are right ! my bad I missed the return on the next line.

mapuri:
LGTM

unclejack:
LGTM

PR #270, opened 2016-05-19
storage/backend/nfs: fix a test that would only break when the VM uses a busted DNS server such as 4.2.2.2.

erikh:


unclejack:
LGTM

PR #272, opened 2016-05-20
Lock free operation

erikh:
This patch implements lock free operation as a toggle. What it does internally is replace the hostname for the exclusive lock on mount to a static string; this way any host can "take" the lock, but it still prevents mounted volumes from existing after a Remove call is issued (it will fail if there are any mounted).

It's not quite ready; there are still some tests to write and some cleanups to do. I was hoping for a preliminary review + CI pass before I move further, as the necessary code is here already.

@mapuri @unclejack @dseevr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ docker-push: docker
 
 run: build
 	vagrant ssh mon0 -c 'volcli global upload < /testdata/globals/global1.json'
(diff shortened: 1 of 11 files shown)


mapuri:
@erikh changes look good but I seem to be missing on the use case. Is this only needed for the nfs backend or it this going to be a more general purpose config driven by user?

erikh:
The NFS situation is different, many containers can take a specific 
volume without consequence, where in the ceph case the RBD devices must 
be locked.

I still need to patch in driver hints, too.

-Erik

unclejack:
The code looks good.

erikh:
I'll look into the failures tomorrow night probably. If not, monday.

erikh:
This is passing and ready for another review, then we can merge it.

mapuri:
LGTM

Issue #247, opened 2016-04-26
dump etcd into a debug-friendly format when requested


erikh:
/cc @jainvipin 

This should have the following changes:

* [x] Respond to SIGUSR2
* [x] Dump etcd keys into /tmp (use mkstemp please)
* [x] tar.gz etcd keys and log filename in volplugin.

erikh:
@unclejack since you are working on other things I'd like to give this ticket to @dseevr who offered to work on it.

unclejack:
@erikh Sure!

dseevr:
This can be closed now 👍

PR #273, opened 2016-05-22
config,volmaster: volmaster handles SIGUSR2 and dumps a gzipped tarball of volplugin's etcd keys

dseevr:
I believe this satisfies all the requirements in #247 

mkstemp is not necessary because ioutil.TempFile implements similar logic using O_EXCL.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,9 +1,17 @@
 package config
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
 	"path"
+	"path/filepath"
+	"time"
 
+	"github.com/contiv/errored"
 	"github.com/contiv/volplugin/watch"
 	"github.com/coreos/etcd/client"
 	"golang.org/x/net/context"
@@ -80,6 +88,93 @@ func (c *Client) prefixed(strs ...string) string {
 	}
 
 	return str
+}
+
+func addNodeToTarball(node *client.Node, writer *tar.Writer, baseDirectory string) error {
+	now := time.Now()
+
+	header := &tar.Header{
+		AccessTime: now,
+		ChangeTime: now,
+		ModTime:    now,
+		Name:       baseDirectory + node.Key,
+	}
+
+	if node.Dir {
+		header.Mode = 0700
+		header.Typeflag = tar.TypeDir
+	} else {
+		header.Mode = 0600
+		header.Size = int64(len(node.Value))
+		header.Typeflag = tar.TypeReg
+	}
+
+	err := writer.WriteHeader(header)
+	if err != nil {
+		return errored.Errorf("Failed to write tar entry header: %v", err)
+	}
+
+	// we don't have to write anything for directories except the header
+	if !node.Dir {
+		_, err = writer.Write([]byte(node.Value))
+		if err != nil {
+			return errored.Errorf("Failed to write tar entry: %v", err)
+		}
+	}
+
+	for _, n := range node.Nodes {
+		addNodeToTarball(n, writer, baseDirectory)
+	}
+
+	return nil
+}
+
+// DumpTarball dumps all the keys under the current etcd prefix into a
+// gzip'd tarball'd directory-based representation of the namespace.
+func (c *Client) DumpTarball() (string, error) {
+	resp, err := c.etcdClient.Get(context.Background(), c.prefix, &client.GetOptions{Sort: true, Recursive: true, Quorum: true})
+	if err != nil {
+		return "", errored.Errorf(`Failed to recursively GET "%s" namespace from etcd: %v`, c.prefix, err)
+	}
+
+	now := time.Now()
+
+	// tar hangs during unpacking if the base directory has colons in it
+	// unless --force-local is specified, so use the simpler "%Y%m%d-%H%M%S".
+	niceTimeFormat := fmt.Sprintf("%d%02d%02d-%02d%02d%02d",
+		now.Year(), now.Month(), now.Day(),
+		now.Hour(), now.Minute(), now.Second())
+
+	file, err := ioutil.TempFile("", "etcd_dump_"+niceTimeFormat+"_")
+	if err != nil {
+		return "", errored.Errorf("Failed to create tempfile: %v", err)
+	}
+	defer file.Close()
+
+	// create a gzipped, tarball writer which cleans up after itself
+	gzipWriter := gzip.NewWriter(file)
+	defer gzipWriter.Close()
+
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
+
+	// ensure that the tarball extracts to a folder with the same name as the tarball
+	baseDirectory := filepath.Base(file.Name())
+
+	err = addNodeToTarball(resp.Node, tarWriter, baseDirectory)
+	if err != nil {
+		return "", err
+	}
+
+	// give the file a more fitting name
+	newFilename := file.Name() + ".tar.gz"
+
+	err = os.Rename(file.Name(), newFilename)
+	if err != nil {
+		return "", err
+	}
+
+	return newFilename, nil
 }
 
 // NotFound is a predicate to determine whether or not the error returned by
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,11 +1,22 @@
 package config
 
 import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"io"
+	"io/ioutil"
+	"os"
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -82,6 +91,98 @@ func (c *Client) prefixed(strs ...string) string {
...
+
+	// create a gzipped, tarballed, buffered writer which self-closes
+	gzipWriter := gzip.NewWriter(file)
+	defer gzipWriter.Close()
+
+	bufWriter := bufio.NewWriter(gzipWriter)
+	defer bufWriter.Flush()
+
+	tarWriter := tar.NewWriter(bufWriter)
+	defer tarWriter.Close()
```
erikh: Hmm. How do you feel about a single `defer` that encapsulates all the close/flush functionality?

I think this will be cleaner because there will be one defer and it will be less confusing which one fires first.

Alternatively, please comment it if you object.

I have no strong opinion, I just wanted to offer an idea.

erikh: Comment the code, rather

config/config.go
```diff
@@ -82,6 +91,98 @@ func (c *Client) prefixed(strs ...string) string {
...
+
+	// ensure that the tarball extracts to a folder with the same name as the tarball
+	parts := strings.Split(file.Name(), "/")
```
erikh: Please use the `path/filepath` and be a little more cautious about path handling, please. volplugin runs as root.

config/config.go
```diff
@@ -82,6 +91,98 @@ func (c *Client) prefixed(strs ...string) string {
...
+
+	err = writeNode(resp.Node, tarWriter, baseDirectoryName)
+	if err != nil {
+		return "", err
+	}
+
+	// give the file a more fitting name
+	newFilename := file.Name() + ".tar.gz"
+
+	err = os.Rename(file.Name(), newFilename)
```
erikh: I think here is another case where we should be very careful about the paths.

config/config_test.go
```diff
@@ -35,3 +42,56 @@ func (s *configSuite) TestPrefixed(c *C) {
 	c.Assert(s.tlc.policy("quux"), Equals, s.tlc.prefixed(rootPolicy, "quux"))
 	c.Assert(s.tlc.volume("foo", "bar", "quux"), Equals, s.tlc.prefixed(rootVolume, "foo", "bar", "quux"))
 }
+
+func (s *configSuite) TestDumpTarball(c *C) {
+	key := "/volplugin/foo"
+	value := "baz"
+
+	exec.Command("/bin/sh", "-c", "etcdctl set "+key+" "+value).Run()
+	defer exec.Command("/bin/sh", "-c", "etcdctl rm "+key).Run()
```
erikh: we have a custom package for this called `executor`. It's not a huge deal, but I'd prefer you use it. This is not mandatory.

https://godoc.org/github.com/contiv/executor

config/config_test.go
```diff
@@ -35,3 +42,56 @@ func (s *configSuite) TestPrefixed(c *C) {
...
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		c.Assert(err, IsNil)
+
+		if header.Name == testFilename {
+			var b bytes.Buffer
+			_, err = io.Copy(&b, tarReader)
+			c.Assert(err, IsNil)
```
erikh: check deliberately for `io.EOF` here.

```go
c.Assert(err, Equals, io.EOF)
```

erikh:
I would prefer that you address the path handling especially, but the other optional tasks; feel free to argue for/against as you have limited time.

LGTM otherwise.

/cc @mapuri @unclejack

config/config_test.go
```diff
@@ -35,3 +42,56 @@ func (s *configSuite) TestPrefixed(c *C) {
 	c.Assert(s.tlc.policy("quux"), Equals, s.tlc.prefixed(rootPolicy, "quux"))
 	c.Assert(s.tlc.volume("foo", "bar", "quux"), Equals, s.tlc.prefixed(rootVolume, "foo", "bar", "quux"))
 }
+
+func (s *configSuite) TestDumpTarball(c *C) {
+	key := "/volplugin/foo"
+	value := "baz"
+
+	exec.Command("/bin/sh", "-c", "etcdctl set "+key+" "+value).Run()
```
unclejack: It might be a good idea to split the code into two separate functions, one which goes and fetches the data from etcd and one which produces the tarball. This would make it easier to add more tests and it'd also make it better when it comes to adding passing/failing test cases.

mapuri:
LGTM other than the comments that are already posted.

config/config_test.go
dseevr: According to [the docs][1]:

```
A successful Copy returns err == nil, not err == EOF. Because Copy is defined to read from src until EOF, it does not treat an EOF from Read as an error to be reported.
```

[1]: https://golang.org/pkg/io/#Copy

config/config_test.go
dseevr: I did see executor, but the `SetUpTest()` function in this file just runs `exec.Command` so I copied that.  Should I update that to use executor as well?

config/config.go
dseevr: Hmm, I've always read/seen that you should be defer'ing cleanup function calls/logic immediately after the setup is complete because a subsequent panic (e.g., y() panics after x() runs but not before x()'s deferred cleanup function is scheduled) would not result in everything being cleaned up properly during stack unwinding to the last recover() call.

defers are always LIFO according to the language spec.  The tar writer writing to the bufio writer is closed (implicit flush into the bufio writer), bufio writer is flushed into the gzip writer (it has no Close() function), gzip writer is closed (implicit flush into the file handle), and then the file itself is closed.

I'm happy to change this if you'd like, but it seems really straightforward to me in its current state.  I also don't think it's a big deal to put all the defers into a single function since it's unlikely that the NewWriter() calls are going to panic unless the system is out of memory, and the LIFO order would be identical anyways.

config/config.go
dseevr: Awesome, I will use this moving forward.

config/config_test.go
dseevr: Thanks for the comment.  Do you really think this is worthwhile?  The code to retrieve the keyspace from etcd is just a [single function call + error check][1], so it doesn't seem like it would really be beneficial to me at this point.

[1]: https://github.com/contiv/volplugin/pull/273/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR136

erikh: I didn't realize this was a test when I first saw this comment. If this was "real code" I would argue with @unclejack , but because this is a test I don't personally care too much.

Let's give @unclejack some time to respond before we merge this. however, I would suggest either preparing an example or doing nothing for now.

config/config_test.go
erikh: maybe I just forgot haha. please disregard.

config/config_test.go
erikh: nah, just leave it be.

config/config.go
erikh: Hmm. How do you feel about this @unclejack @mapuri?

config/config_test.go
unclejack: After seeing the test, I figured out this would be a good idea to make it easier to test without relying on etcd.

erikh:
@unclejack works nights our time, and I've asked him a bunch of questions so I'm not going to merge this today, but I suspect there will be no problems tomorrow.

Thank you for both the patch and the engaging discussions.

Once path handling is done I think we're good to go for me.

config/config_test.go
erikh: I understand but as we discussed elsewhere I suggested that this doesn't need to be changed because it's only used in one place. gocheck will output the code that errors so we should have a good idea without a new function to determine that it's this code.

FWIW, I am going to suggest we not implement this for now.

config/config.go
mapuri: personally i like the current way better as it keeps the cleanup code close to setup code, so it's easy to verify that we didn't miss on a Close() or a Flush(), so this looks good to me.

erikh: Ok, I won't argue this direction any further. Sorry to waste anyone's time.

config/config_test.go
unclejack: That's fine.

unclejack:
Once the pending changes are made, LGTM.

config/config.go
dseevr: Discussed this with @erikh, this is a straight up append of the ioutil.TempFIle() output so there's really no way to do this in a safer manner than it's currently done.

erikh:
Thanks for your contribution!

PR #274, opened 2016-05-23
add version support in build target

mapuri:
this let's us pass the version string from release CI to the built binaries thereby tying the binaries downloaded from github to respective releases.

Also it enables us to push versioned volplugin container images (with every github release) that we can use for setting up volplugin in our ansible. This will be done is a subsequent PR.

/cc @erikh @dseevr @unclejack


mapuri:
closing as something seems wrong with the branch cause merge checks to fail in github. Will file another PR

PR #275, opened 2016-05-23
add version support in build target

mapuri:
this let's us pass the version string from release CI to the built binaries thereby tying the binaries downloaded from github to respective releases.

Also it enables us to push versioned volplugin container images (with every github release) that we can use for setting up volplugin in our ansible. This will be done is a subsequent PR.

/cc @erikh @dseevr @unclejack

### DIFF
--- a/volmaster/volmaster/cli.go
+++ b/volmaster/volmaster/cli.go
@@ -12,6 +12,9 @@ import (
 	"github.com/codegangsta/cli"
 )
 
+// version is provided by build
+var version = ""
+
 func start(ctx *cli.Context) {
 	cfg, err := config.NewClient(ctx.String("prefix"), ctx.StringSlice("etcd"))
 	if err != nil {
@@ -29,7 +32,7 @@ func start(ctx *cli.Context) {
 
 func main() {
 	app := cli.NewApp()
-	app.Version = ""
+	app.Version = version
 	app.Usage = "Control many volplugins"
 	app.Action = start
 	app.Flags = []cli.Flag{
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,9 @@ run-volmaster:
 	sudo -E nohup bash -c '$(GUESTBINPATH)/volmaster &>/tmp/volmaster.log &'
 
 run-build: 
-	GOGC=1000 go install -v ./volcli/volcli/ ./volplugin/volplugin/ ./volmaster/volmaster/ ./volsupervisor/volsupervisor/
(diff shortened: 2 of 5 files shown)


volmaster/volmaster/cli.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/codegangsta/cli"
 )
 
+// version is provided by build
+var version = ""
```
erikh: Think we need to do a dockerversion/netctl version like thing here? I can do it in another patch, but it'll kill the `go get` notion.

erikh:
code LGTM but I don't see any versioning controls beyond renaming the file. Is that the intention?

volmaster/volmaster/cli.go
mapuri: @erikh , this PR will bring in the equivalent of  `docker --version`, which is much smaller and local in scope but is useful for tying our binaries to respective releases.

 `docker version` and `netctl version` have a larger and distributed scope of having a REST endpoint to deal with various distributed components of a system. We should perhaps debate it's need in a separate issue.

mapuri:
>  I don't see any versioning controls beyond renaming the file. Is that the intention?

Yes, I set the version of built binaries to be same as our existing release versions (i.e. version string is generated and passed to `make build` as part of the `make release` target), so this doesn't need an explicit version control.

erikh:
I don't really like not having internal control of versioning; seems 
like we'll bring pain as far as independent releases go. Let's back out 
the string changes; I will come up with something. Let's focus on 
renaming the file in the make task for now.

mapuri:
hey @erikh 

> Let's focus on renaming the file in the make task for now.

there is no file rename related change in this PR. The changes just define a `main.version` variable and pass it's value as part of the `make build` using `-ldFlags'. That is the whole PR, if we revert that then nothing else will be left to commit, unless I misread your comment :)

As for independent releases we can control it from CI job (i.e. we control the release versions from there), something that do for packer builds but that can be easily built on top of this change.

erikh:
Ohhh I see what this is doing now. Very sorry. Thank you for your contribution, LGTM.

erikh:
Honestly, now that I see what it's doing; I am very excited about this patch. Nice find.

volmaster/volmaster/cli.go
erikh: Well, we'll be removing the volmaster after release and rewriting it as 
an API service only, so that will be the time to consider it.

-Erik

Issue #229, opened 2016-04-15
volmaster REST endpoints need some rework


mapuri:
tracking the items from comments here: https://github.com/contiv/volplugin/pull/228/files#r59947447

(edited by @erikh)

* [x] for everything that uses etcd straight from volcli, please implement the appropriate REST call in the volmaster instead, and point volcli to use it.
* [x] Fix routes to be clearer (covered by #255)
* [x] Clean up existing endpoints if necessary -- structs in use should be typed structs and not f.e. maps.
* [x] Normalize http.Client usage so that a shared client is used.
* [ ] Remove all `fmt.Sprintf` calls that generate URLs and use a URL constructor instead (encoding/url maybe?)
* [x] Make it possible to get mount information for volumes (implemented by #280)
* [x] Make it possible to get mount information for snapshots
* [x] Switch remove handlers to DELETE (implemented by #280 already)

mapuri:
List of work items:
- [ ] make GET endpoints consistent - will be addressed by #255 
- [ ]  make POST endpoints consistent

erikh:
Closing this; it looks like we're ready to go with these patches and they made it into the v0.1 release. We will revisit the client issue in a separate ticket.

PR #280, opened 2016-05-26
more changes for the routes

unclejack:
Makes changes needed for #229.

### DIFF
--- a/volcli/volcli.go
+++ b/volcli/volcli.go
@@ -56,6 +56,16 @@ func ppJSON(v interface{}) ([]byte, error) {
 	return json.MarshalIndent(v, "", "  ")
 }
 
+func deleteRequest(url string, bodyType string, body io.Reader) (resp *http.Response, err error) {
+	req, err := http.NewRequest("DELETE", url, body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare request: %v", err)
+	}
+	req.Header.Set("Content-Type", bodyType)
+
+	return http.DefaultClient.Do(req)
+}
+
 // GlobalGet retrives the global configuration and displays it on standard output.
 func GlobalGet(ctx *cli.Context) {
 	execCliAndExit(ctx, globalGet)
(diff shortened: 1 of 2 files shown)


volmaster/daemon.go
```diff
@@ -90,15 +90,26 @@ func (d *DaemonConfig) Daemon(listen string) {
 		r.HandleFunc(path, logHandler(path, d.Global.Debug, f)).Methods("POST")
 	}
 
+	deleteRouter := map[string]func(http.ResponseWriter, *http.Request){
+		"/volumes/remove":           d.handleRemove,
```
mapuri: @unclejack we should remove these routes from POST router above, right?

mapuri:
LGTM, except for one minor clarification.

volmaster/daemon.go
unclejack: @mapuri You're right. I'll update right away.

volcli/volcli.go
```diff
@@ -56,6 +56,15 @@ func ppJSON(v interface{}) ([]byte, error) {
 	return json.MarshalIndent(v, "", "  ")
 }
 
+func deleteRequest(url string, bodyType string, body io.Reader) (resp *http.Response, err error) {
```
erikh: why are we passing bodyType without using it?

volmaster/daemon.go
```diff
@@ -220,6 +228,40 @@ func (d *DaemonConfig) handlePolicy(w http.ResponseWriter, r *http.Request) {
...
+
+	_, err := d.Config.GetVolume(policy, volumeName)
+	if config.NotFound(err) {
+		w.WriteHeader(404)
+		return
+	} else if err != nil {
+		httpError(w, "Retrieving volume", err)
+	}
+
+	if err := d.Config.GetUse(mount, vc); err != nil {
```
erikh: Is the volume fetch necessary here? I'm not sure it is.

erikh:
looks good otherwise

volcli/volcli.go
unclejack: I'll fix it.

unclejack:
@erikh The fixes are in and the CI is green. PTAL

volcli/volcli.go
erikh: this is still needed

erikh:
looks like we have one last issue here

volcli/volcli.go
unclejack: @erikh It's used below to set "Content-Type". That's the fix I've made.

erikh:
aha! LGTM

PR #283, opened 2016-05-29
Make locking great again

### LINKED ISSUE #282: volume mounted multiple times
Hello,

I tested the plugin with make demo and tried the scenario to mount the same volume multiple times.

What I noticed is that the first time I try to mount a volume which is already mounted somewhere else it fails with message 
```
docker: Error response from daemon: VolumeDriver.Mount: Mountpoint already in use <nil>.
```
The second time I try though the volume mounts normally and I could write from both containers and read from them.
I tested on the same host and did not try running the container in a different host.

erikh:
Fixes a number of HTTP handling errors in the volmaster. The other fixes are being sent to CI to ensure they are compatible fixes.

Will replace this with useful context once all the tests, etc are written.

resolves #282


erikh:
Ok, even though this should theoretically pass CI I don't believe it's the solution. Closing this.

PR #286, opened 2016-05-30
Fix locking

### LINKED ISSUE #282: volume mounted multiple times

erikh:
Fixes #282.

This changes the mounting semantics when volumes are mounted for multiple containers on the same host. Previously, volplugin disallowed this, but the code was exceptionally racy.

Now, it allows multiple containers to mount a specific volume on a single host.

/cc @mapuri @unclejack @jainvipin @dseevr


unclejack:
I'm getting the same failure in the tests:
```
----------------------------------------------------------------------
FAIL: battery_test.go:15: systemtestSuite.TestBatteryMultiMountSameHost

battery_test.go:54:
    c.Assert(err, IsNil)
... value *ssh.ExitError = &ssh.ExitError{Waitmsg:ssh.Waitmsg{status:1, signal:"", msg:"", lang:""}} ("Process exited with: 1. Reason was:  ()")
```

This test fails every time with NFS.

erikh:
Yes I'm working on this.

erikh:
This is ready for review. Mind the commits; I had to amend one and add another.

volplugin/utils.go
```diff
@@ -92,7 +92,7 @@ func httpError(w http.ResponseWriter, message string, err error) {
 	}
 
 	log.Warnf("Returning HTTP error handling plugin negotiation: %s", fullError)
-	http.Error(w, string(content), http.StatusInternalServerError)
+	http.Error(w, string(content), http.StatusOK)
```
unclejack: Why is this changed?

volplugin/mountcount.go
```diff
@@ -0,0 +1,17 @@
+package volplugin
+
+func (dc *DaemonConfig) increaseMount(mp string) int {
+	dc.mountMutex.Lock()
+	defer dc.mountMutex.Unlock()
+
+	dc.mountCount[mp]++
```
unclejack: This state seems to not be stored across daemon restarts.

I've commented to make sure this is tracked.

volplugin/utils.go
erikh: I'll back that out into another commit.

erikh:
Don't review this yet. Sorry for the flip-floppiness.

erikh:
The fundamental problem here is that docker will have no problem accepting multiple bind mounts to the same place but NFS in particular returns EBUSY on a double-mount (which makes complete sense).

In this patch is a reference counter for mounts on the same host, so that multiple mounts do not perform the actual mount actions, and therefore it ignores the mount action and records something in-memory.

The problem now, however, is that there's no way to get at a bind mount that's been placed inside a namespace, at least with my limited knowledge of kernel facilities. Therefore, when volplugin boots it can only record one mount, and the first mount request will unmount it.

So, the solution here is really to take the mount recording in the storage drivers and pair it with docker requests that search for bind mounts for our plugin.

I'll be doing this in future patches to this PR. Please let me know if you disagree with this approach for some reason.

mapuri:
@erikh , overall the plan sounds good. I have a few clarifications for my understanding.

> The fundamental problem here is that docker will have no problem accepting multiple bind mounts to the same place but NFS in particular returns EBUSY on a double-mount (which makes complete sense).

will this happen even if the user were to use an existing NFS volume on the host inside docker run **without** the volplugin i.e. would the following fail as well? If so, then we are actually matching native behavior, right? Do we need to change it with volplugin?

```
docker run -v /existing/nfs/volumeA:/inside/volume1 image1
docker run -v /existing/nfs/volumeA:/inside/volume2 image2
```

> therefore it ignores the mount action and records something in-memory.

I am not sure if I completely follow the implications of doing this. Is the volume still mounted in the second container? If not, then ignoring the mount could break the app in the container, right?

> So, the solution here is really to take the mount recording in the storage drivers and pair it with docker requests that search for bind mounts for our plugin.

yes, this sounds good to me. I think docker at some point was planning to give UUIDs per volume per container. May be we can use it to keep this map when it is available?

erikh:
This should be ready for review, now.

erikh:
@mapuri to respond to your questions:

volplugin is now scanning the running containers to see if any mounts are both belonging to volplugin and have a name that matches one of our volumes. It then increments the mount counter. This is done at init time, but no later attempt to reconcile state is required (since this is all managed only for the runtime session). This is working and probably acceptable even for large container workloads since docker's `ps` does the same thing.

Re: UUID, that was a change I suggested to cpuguy that won't be accurate for each volume name I think, but instead per-volume per-container, which makes our logic much more complicated. After suggesting this, I don't think we'll actually use it haha.

erikh:
Looking into the failures.

erikh:
Making another attempt; I spent an hour running tests in a loop and cannot reproduce this yet.

erikh:
Free to review now. Thanks for your patience all!

volplugin/mountcount.go
unclejack: What about this?

erikh: yes, read below please.

volplugin/init.go
```diff
@@ -0,0 +1,92 @@
...
+	log "github.com/Sirupsen/logrus"
+)
+
+func (dc *DaemonConfig) updateMounts() error {
+	dockerClient, err := client.NewEnvClient()
+	if err != nil {
+		return errored.Errorf("Could not initiate docker client").Combine(err)
+	}
+
+	containers, err := dockerClient.ContainerList(context.Background(), types.ContainerListOptions{})
```
mapuri: do we need to worry about containers that are down or went down while volplugin was down?

mapuri:
LGTM, except for one clarification.

volplugin/init.go
erikh: Theoretically yes, but I'm not sure how I could reconcile that.

erikh: Please let me know if you have any suggestions! I don't believe I considered this case before.

dseevr:
LGTM

erikh:
Gonna wait for @mapuri's signoff before I move forward since he had some outstanding Q&A.

unclejack:
LGTM

erikh:
@mapuri if you could detail your concerns in a separate issue, I will examine and patch it in a separate PR. Going to merge this.

mapuri:
nah, this LGTM @erikh thanks!

PR #288, opened 2016-05-31
volci,volmaster: drop DELETE from DELETE URLs

unclejack:


mapuri:
@unclejack LGTM

I think at some point we will need to make `/volumes/remove` and `/volumes/removeforce` consistent with this naming as well.

erikh:
LGTM. Is this ready for merge?

erikh:
(after CI runs)

unclejack:
@erikh Yes, it should be good.

@mapuri: Sure, it's just that this was a smaller change to make.

PR #289, opened 2016-06-01
config,volcli,volmaster: send full policy JSON

unclejack:
This PR changes the REST api for the policy listing to send an array of complete policies.

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"encoding/json"
-	"path"
 
 	"github.com/contiv/errored"
 	"github.com/contiv/volplugin/storage/backend/ceph"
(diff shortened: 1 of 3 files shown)


volmaster/daemon.go
```diff
@@ -203,7 +203,12 @@ func (d *DaemonConfig) handlePolicyList(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	content, err := json.Marshal(policies)
+	response := []*config.Policy{}
+	for _, policy := range policies {
+		response = append(response, policy)
```
erikh: why not just marshal policies directly here exactly?

erikh:
LGTM, but I think the policies loop is superfluous, unless there's good reason, please remove it.

volmaster/daemon.go
unclejack: Fair enough, I'll refactor the code and update.

unclejack:
@erikh PTAL

dseevr:
LGTM

erikh:
LGTM

Issue #278, opened 2016-05-24
Creating volume Formatting volume: Error creating filesystem on /dev/rbd0 with cmd: \"mkfs.btrfs %\". Error: Command: [/bin/sh -c mkfs.btrfs /dev/rbd0], Exit status 1


angry-tony:
Hi.. i'm new to volplugin.
when i ran "docker volume create" to use btrfs for docker volume, i got this error.
but ext4 is ok.
i guess btrfs's options need to change.

```
[cluster-admin@node-01 ~]$ sudo docker volume create -d volplugin --name hoho/foo

[...]

PR #290, opened 2016-06-01
Better errors

### LINKED ISSUE #182: volcli improvement: detect etcd missing keys and respond with nicer messages
Currently this would involve changes to the volmaster, config library and volcli. If the volmaster was smart enough to return 404 on etcd "key not found" failures, it would be a much nicer user experience, and easier to test in the long term I think. It could also exit a different value if the etcd key is missing.

erikh:
This drastically improves the errors that volcil, volplugin and volmaster return. Other modifications will come later, but the pattern and the errors are defined and integrated in the worst pain points.

/cc @jainvipin 

This builds on the same host mount patch, so please review that first.

Closes #182

### DIFF
--- a/vendor/github.com/contiv/errored/README.md
+++ b/vendor/github.com/contiv/errored/README.md
@@ -24,6 +24,7 @@ func main() {
 	combined := err.Combine(err2)
 	combined.SetTrace(true)
 	combined.Error() // => "a message: another message" + two stack traces
+  combined.Contains(err2) // true
 }
 ```
 
--- a/volplugin/init.go
+++ b/volplugin/init.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/contiv/errored"
 	"github.com/contiv/volplugin/config"
+	"github.com/contiv/volplugin/errors"
 	"github.com/contiv/volplugin/lock"
 	"github.com/contiv/volplugin/storage/backend"
 	"github.com/docker/engine-api/client"
@@ -58,12 +59,16 @@ func (dc *DaemonConfig) updateMounts() error {
 			log.Infof("Refreshing existing mount for %q", mount.Volume.Name)
 
 			vol, err := dc.requestVolume(parts[0], parts[1])
-			switch err {
-			case errVolumeNotFound:
-				log.Warnf("Volume %q not found in database, skipping")
-				continue
-			case errVolumeResponse:
-				log.Fatalf("Volmaster could not be contacted; aborting volplugin.")
+			if erd, ok := err.(*errored.Error); ok {
+				switch {
+				case erd.Contains(errors.NotExists):
+					log.Warnf("Volume %q not found in database, skipping")
+					continue
+				case erd.Contains(errors.GetVolume):
+					log.Fatalf("Volmaster could not be contacted; aborting volplugin.")
+				}
+			} else if err != nil {
+				log.Fatalf("Unknown error reading from volmaster: %v", err)
 			}
 
 			payload := &config.UseMount{
--- a/volplugin/utils.go
+++ b/volplugin/utils.go
@@ -3,7 +3,6 @@ package volplugin
 import (
 	"bytes"
 	"encoding/json"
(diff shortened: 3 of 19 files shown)


volplugin/utils.go
```diff
@@ -71,28 +66,26 @@ func (dc *DaemonConfig) structsVolumeName(uc *unmarshalledConfig) (storage.Mount
...
 	if errc != nil {
-		log.Warnf("Error received marshalling error response: %v, original error: %s", errc, fullError)
+		log.Warnf("Error received marshalling error response: %v, original error: %s", errc, err.Error())
 		return
 	}
 
-	log.Warnf("Returning HTTP error handling plugin negotiation: %s", fullError)
-	http.Error(w, string(content), http.StatusInternalServerError)
+	log.Warnf("Returning HTTP error handling plugin negotiation: %s", err.Error())
+	http.Error(w, string(content), http.StatusOK)
```
dseevr: Why was the status code here changed from 500 to 200?

dseevr:
LGTM aside from that one issue

volplugin/utils.go
erikh: Docker doesn't actually support HTTP status codes, so I corrected it, basically.

erikh:
Please review @unclejack @mapuri

vendor/github.com/contiv/errored/README.md
```diff
@@ -24,6 +24,7 @@ func main() {
 	combined := err.Combine(err2)
 	combined.SetTrace(true)
 	combined.Error() // => "a message: another message" + two stack traces
+  combined.Contains(err2) // true
```
mapuri: the indentation seems off here

volplugin/init.go
```diff
@@ -58,12 +59,16 @@ func (dc *DaemonConfig) updateMounts() error {
...
-				continue
-			case errVolumeResponse:
-				log.Fatalf("Volmaster could not be contacted; aborting volplugin.")
+			if erd, ok := err.(*errored.Error); ok {
+				switch {
+				case erd.Contains(errors.NotExists):
+					log.Warnf("Volume %q not found in database, skipping")
+					continue
+				case erd.Contains(errors.GetVolume):
+					log.Fatalf("Volmaster could not be contacted; aborting volplugin.")
```
mapuri: this is minor and probably not in the scope of this PR but thought will share and see if we need to handle this.

Just thinking out loud, for cases like these should we just wait and retry like we do for etcd as volmaster might not be up when volplugin starts

volplugin/utils.go
```diff
@@ -71,28 +66,26 @@ func (dc *DaemonConfig) structsVolumeName(uc *unmarshalledConfig) (storage.Mount
...
-func httpError(w http.ResponseWriter, message string, err error) {
-	fullError := fmt.Sprintf("%s %v", message, err)
-
-	content, errc := marshalResponse(VolumeResponse{"", fullError})
+func httpError(w http.ResponseWriter, err error) {
+	content, errc := marshalResponse(VolumeResponse{"", err.Error()})
 	if errc != nil {
-		log.Warnf("Error received marshalling error response: %v, original error: %s", errc, fullError)
+		log.Warnf("Error received marshalling error response: %v, original error: %s", errc, err.Error())
 		return
```
mapuri: This is not in scope of this PR but may be this `return` should be replaced by setting `content = ""` otherwise we would not respond to the http request at all?

mapuri:
LGTM , pretty nice work!

volplugin/init.go
erikh: This is addressed before this code is reached in the `Daemon()` call.

vendor/github.com/contiv/errored/README.md
erikh: this is in `errored`, so I won't address it here.

erikh:
Thanks! I'll merge after tests pass.

erikh:
Fixes #278

PR #291, opened 2016-06-01
Units

erikh:
This removes our dependency on alecthomas/units. I removed this package because `docker/go-types` is more flexible and in use by many of our dependencies.

other tickets need to be merged first.


erikh:
This should be ready for review

mapuri:
LGTM

PR #293, opened 2016-06-03
volmaster: add route to fetch snapshot mounts

unclejack:


erikh:
I have this covered in #299 now. We can probably close it as there is volcli integration over there too.

PR #301, opened 2016-06-09
Fixes for same-host multi-mount mode

erikh:
This fixes a bug in the mount counting code that would cause it to go negative. You would have to acquire volumes very fast on multiple hosts and then it to happen, which is a significant edge case.

To prevent future bugs in the future like this, I have added some code to panic the daemon if the mount count goes negative. We can do serious damage if we don't, as the counting system gets off and duplicate maps/mounts happen.


erikh:
This is ready for review.

mapuri:
LGTM

PR #302, opened 2016-06-10
vendor the recent ansible repo and do respective fixes in systemtests

mapuri:
systemtests fixes to bring in new ansible involve:
- clean up only the containers started by the tests
  - In this patch I did it by tracking the `name`s of started container. 
- also added a wait for etcd cluster to be formed whenever docker is restarted
  - it might improve test times if we can remove certain docker restarts, but I haven't evaluated that in this PR

/cc @erikh. 

I am still running tests locally as well but as they take some time thought I will also share the approach to get early feedback.


erikh:
looks GREAT! Let's wait for the PR to finish checking, then merge.

Thank you so much!

mapuri:
@erikh this is ready for review. The changes from the last patch are:
- safeguarded writes to the container's name list variable `startedContainers` within locks (to ensure correct updates to the data structure by multiple go-routines)
- removed a docker restart that used to happen in every iteration of `TestBatteryParallelMount` test. This causes etcd to be restarted on all nodes causing etcd data (and hence policy data) to be lost. 

PTAL when you get a chance and let me know your comments.

erikh:
This looks good. Do we restart docker anywhere else that might need to be removed?

erikh:
looks like ansible failed because of a download; can you re-push?

mapuri:
@erikh yes, I just pushed again.

Also I tried with docker restart removed from rebootstrap but I am seeing some sort of flakiness in the tests with it that I am not able to explain. For instance http://contiv.ngrok.io/view/Volplugin%20CI/job/Volplugin_CI/723/console is the failed job with docker restart removed from rebootstrap. The failure makes me feel that we might be missing some sort of other cleanup  between tests that docker restart might be helping with but I couldn't find anything obvious.

For now I am trying again with this particular change reflected to see if it goes through reliably.

erikh:
I know what this is; feel free to revert. It's a docker bug :(

erikh:
Looks like it's still failing. Maybe a locking bug (I'm seeing one for NFS in my own PRs) but the remove errors before seem to indicate it's not.

erikh:
I know you're not done but what you're doing is amazing :)

mapuri:
@erikh this patch has passed sanity a couple of times, so I think it is good for another look.

Basically `TestBatteryParallelMount` test runs into intermittent failures if we don't restart docker after every loop, so to work-around it I have kept the docker restart as is and instead added a wait for etcd to come up and also a policy push to create necessary state for each loop. Note that this has resulted in increased the sanity run time as a side effect.

PTAL when you get a chance.

erikh:
So, the fundamental cause is with the create operation. If the volume is removed out of band it takes a while before docker figures it out during create, even though it attempts to gather information about an empty volume. However, at this time the create operation *succeeds* instead of fails, which gives us a false-positive about the state of things as we proceed.

I'm not sure what the fix is here and it very well could be our problem, so I don't want to run over to the tracker until I have a full story.

erikh:
LGTM. Thank you so much for your help.

PR #310, opened 2016-06-15
Global driver

erikh:
This implements the necessary API calls to present ourselves as a global driver to docker.

Necessary in Swarm 1.1 and up, and docker 1.12.


erikh:
This also contains a necessary fix to demo/start modes which would improperly provision the last host at `make run` time. See comments in the commit message.

mapuri:
LGTM

PR #313, opened 2016-06-21
Volmaster refactor

erikh:
This moves all locking, creation, and even a few API calls out of volmaster into separate libraries, and those libraries are then used by both volmaster/volplugin. This only covers a few calls now, but I intend to extend it with more once we rebuild the API service from the volmaster.

Note that I still have a lot of tests here, but the core functionality patch-wise is done. Please take your time with this review, set aside an hour or two.

/cc @mapuri @dseevr @unclejack


lock/lock.go
```diff
@@ -110,3 +102,92 @@ func (d *Driver) ExecuteWithMultiUseLock(ucs []config.UseLocker, timeout time.Du
...
+	return stopChan, nil
+}
+
+func (d *Driver) lockWait(uc config.UseLocker, timeout time.Duration, now time.Time, reason string) (bool, error) {
+	log.Warnf("Could not %s %q lock for %q", reason, uc.GetReason(), uc.GetVolume())
+	if timeout != 0 && (timeout == -1 || time.Now().Sub(now) < timeout) {
+		log.Warnf("Waiting 100ms for %q lock on %q to free", uc.GetReason(), uc.GetVolume())
+		time.Sleep(wait.Jitter(100*time.Millisecond, 0))
+		return true, nil
+	} else if time.Now().Sub(now) >= timeout {
```
dseevr: `time.Since(now)` is a [shorthand way to do this](https://golang.org/pkg/time/#Since)

api/api.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/contiv/volplugin/errors"
+	"github.com/contiv/volplugin/lock"
+	"github.com/contiv/volplugin/storage/control"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// VolumeRequest is taken from
+// https://github.com/calavera/docker-volume-api/blob/master/api.go#L23
+type VolumeRequest struct {
```
mapuri: I think docker has these structures defined as well (https://github.com/docker/docker/blob/master/vendor/src/github.com/docker/engine-api/types/types.go#L437), may be we should use those or vendor them in to avoid any incompatibilities between structs.

mapuri:
LGTM, apart from one minor clarification.

api/api.go
erikh: Ah ok; I will retool to use those then.

erikh: putting in a ticket for this one, will address at a later point.

volcli/commands.go
```diff
@@ -89,7 +89,7 @@ var Commands = []cli.Command{
 					Usage: "Provide key=value options to create the volume",
```
unclejack: Are there any use commands we should be getting rid of now?

volplugin/runtime.go
```diff
@@ -1,96 +1,52 @@
...
-			log.Debugf("Checking runtime parameters for volume %q (ttl: %d)", volumeName, dc.Global.TTL)
+		volWatch := <-volumeChan
 
-			log.Debugf("Requesting runtime configuration for volume %q", volumeName)
-			resp, err := http.Get(fmt.Sprintf("http://%s/runtime/%s", dc.Master, volumeName))
-			if err != nil {
-				log.Error(errored.Errorf("Error processing runtime update").Combine(err))
-				continue
-			}
+		fmt.Printf("%#v\n", volWatch)
```
unclejack: Should this be in here?

volcli/commands.go
erikh: No.

volplugin/runtime.go
erikh: No. :D

dseevr:
LGTM

erikh:
will be merging this once tests pass.

PR #320, opened 2016-06-29
api,volplugin: code refactor - isolation of volume structs

### LINKED ISSUE #317: Correct plugin code to use docker types
These types weren't public before when docker was first supporting volume plugins, but now they are. See the corresponding discussion.

https://github.com/contiv/volplugin/pull/313#discussion_r68489642

yuva29:
Implementation for #317  

fixes #317

### DIFF
--- a/docker/structs.go
+++ b/docker/structs.go
@@ -1,8 +1,0 @@
-package docker
-
-import "github.com/docker/engine-api/types"
-
-// VolumeRequest is taken from
-// https://github.com/docker/engine-api/types/types.go
-type VolumeRequest types.VolumeCreateRequest
-
--- a/docker/structs.go
+++ b/docker/structs.go
@@ -1,0 +1,8 @@
+package docker
+
+import "github.com/docker/engine-api/types"
+
+// VolumeRequest is taken from
+// https://github.com/docker/engine-api/types/types.go
+type VolumeRequest types.VolumeCreateRequest
+
--- a/api/api.go
+++ b/api/api.go
@@ -12,16 +12,10 @@ import (
 	"github.com/contiv/volplugin/errors"
 	"github.com/contiv/volplugin/lock"
 	"github.com/contiv/volplugin/storage/control"
+	"github.com/contiv/volplugin/docker"
 
 	log "github.com/Sirupsen/logrus"
 )
-
-// VolumeRequest is taken from
-// https://github.com/calavera/docker-volume-api/blob/master/api.go#L23
-type VolumeRequest struct {
-	Name string
-	Opts map[string]string
-}
 
 // VolumeResponse is taken from
 // https://github.com/calavera/docker-volume-api/blob/master/api.go#L23
@@ -50,7 +44,7 @@ func (a *API) Create(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var req VolumeRequest
+	var req docker.VolumeRequest
 
 	if err := json.Unmarshal(content, &req); err != nil {
 		a.HTTPError(w, errors.UnmarshalRequest.Combine(err))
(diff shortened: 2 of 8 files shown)


docker/structs.go
```diff
@@ -0,0 +1,8 @@
+package docker
+
+import "github.com/docker/engine-api/types"
+
+// VolumeRequest is taken from
+// https://github.com/docker/engine-api/types/types.go
+type VolumeRequest types.VolumeCreateRequest
```
mapuri: just curious is there a reason to redefine the `VolumeCreateRequest` type? We could just use `types.VolumeCreateRequest` everywhere, right? 

I would prefer the latter as that's what we do elsewhere in the project. 

If we decide to do the current way, it might be better to keep this type in api package itself i.e. here https://github.com/contiv/volplugin/pull/320/files#diff-604f4192deea5adf9c159e5f88f64630L21 , that will limit the change just to one file

mapuri:
@yuva29 looks like a good start ! I have one comment that I shared above also checkout the docker's volume api tests https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/integration-cli/docker_api_volumes_test.go

I think we will need to bring in the types for response as well which might be slightly bigger effort. I will let @erikh chime in too as this might be just me :)

docker/structs.go
erikh: I wanted a type mapping because docker's API is still extremely volatile, and not all structs from the volumes API are represented yet.

This gets compiled away so it should be performant. I asked her to put it in `docker/` but I think you have a good point about putting it in `api/` instead. @yuva29 do you have any objections to doing that?

yuva29: @mapuri Makes sense. I can move them to the same file as it was before. But just a thought, wouldn't it be better to isolate all the volume related structures in a single file. I see that struct 'Volume' is defined in a different file and 'volumeGetRequest' in some other. Please let me know your thoughts @mapuri @erikh

erikh: Please keep them in a separate file, just no need for a new package.

erikh: Can you link me to the relevant types in engine-api so I can make an informed decision?

mapuri: hmmm, the more I read docker remote volume plugin code the more it looks like that none of the volume RPC structures are public.

For instance looks like the plugin RPC structures are generated using the [plugin-rpc tool] (https://github.com/docker/docker/blob/28dde09cdf2a48dcdb6ce665cd2b912175dfb871/pkg/plugins/pluginrpc-gen/README.md) that works on a [private interface](https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/volume/drivers/extpoint.go#L30) which generates code like [this] (https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/volume/drivers/proxy.go)

Not sure if it's even a good idea to use `types.VolumeCreateRequest` unless I am still missing something. We may need to keep the approach that we have. 

@erikh @yuva29 wdyt?

mapuri: ah, may be I spoke too soon :) 

Here is what docker is proposing to use: https://github.com/docker/go-plugins-helpers/tree/master/volume

yuva29: @mapuri Yes. types.VolumeCreateRequest isn't compatible with VolumeRequest. And, the  build failed because of the structure compatibility. We can stick with the old structure.

@erikh Seems like a lot changed in Response and Request https://github.com/docker/go-plugins-helpers/blob/master/volume/api.go As discussed I'm just moving the structures.

erikh: If we can re-use all the structs here:

* if they cover all the API, no need for the shim
* if they only cover parts of the API, please insert shim.

You'll need to import that dependency; let me know when you're there, I need to show both you and @dseevr how to do this I'm guessing.

erikh: Oh. Please do not use the handlers there.

erikh:
LGTM, but please reformat your commit message to use the format we want in CONTRIBUTING.md.

erikh:
I'm going to merge this as-is. Thanks!

Issue #319, opened 2016-06-27
Staging dev environment on multiple networks


erikh:
The networks inside the Vagrantfile should be parameterized (or randomized, perhaps both) so that they can coexist with other VMs on the same host. This will allow multiple dev envs to coexist on the same host.

PR #321, opened 2016-06-30
Vagrantfile: randomize the third octet of the subnet to allow for up to 255 development environments

dseevr:
Octet reservation is protected by locking so it’s safe to bring up many environments at the same time.

Addresses #319

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,9 +6,9 @@ require 'fileutils'
 VAGRANTFILE_API_VERSION = '2'
 OSX_VMWARE_DIR = "/Applications/VMware Fusion.app/Contents/Library/"
 
-config_file=File.expand_path(File.join(File.dirname(__FILE__), 'vagrant_variables.yml'))
-ansible_config_file=File.expand_path(File.join(File.dirname(__FILE__), 'ansible', 'ansible.cfg'))
-settings=YAML.load_file(config_file)
+config_file = File.expand_path(File.join(File.dirname(__FILE__), 'vagrant_variables.yml'))
+ansible_config_file = File.expand_path(File.join(File.dirname(__FILE__), 'ansible', 'ansible.cfg'))
+settings = YAML.load_file(config_file)
 
 if ENV["DEMO"]
   settings["vms"] = 1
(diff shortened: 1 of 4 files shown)


dseevr:
Not ready for this to merge, just wanted to get CI testing it.  Need to remove debugging info + a bit of other cleanup.

Vagrantfile
```diff
@@ -18,11 +18,96 @@ end
...
+      raise SubnetStateFileError.new(msg.join(" "))
+    end
+  end
+
+  # no existing subnet assignment, so reserve a new one
+
+  FileUtils.touch(SUBNET_STATE_FILE)
+  FileUtils.touch(SUBNET_STATE_LOCKFILE)
+
+  f = File.open(SUBNET_STATE_LOCKFILE)
```
erikh: you're gonna hate me, but you can chuck probably 50% of this code if you just use a directory as your database with the randomized octet as filenames in it. Locking is for free that way.

dseevr: That's actually not a bad idea at all.  Thanks.

Vagrantfile
```diff
@@ -18,17 +18,90 @@ end
...
 
 NMONS        = ENV["VMS"] || settings['vms']
-SUBNET       = settings['subnet']
 BOX          = settings['vagrant_box']
 BOX_VERSION  = settings['box_version']
 memory       = settings['memory']
 
+SUBNET_PREFIX          = settings['subnet_prefix']
+SUBNET_ASSIGNMENT_DIR  = "/tmp/volplugin_vagrant_subnets/"
+SUBNET_ASSIGNMENT_FILE = "subnet_assignment.state"
```
erikh: this file is getting created in the wrong directory sometimes. Please use the pattern on line 9 for calculating the directory.

erikh:
LGTM minus the issue noted above, I tested this and it works great.

For the future, just mkdir + checking for EEXIST and re-rolling with rand will work. No need for the existence checks.

erikh:
ah; that is what you're doing, disregard. :D

Vagrantfile
dseevr: Good catch.  Thanks!

I'm curious:  what other directories would it be created in?  Do we invoke vagrant from some other CWD sometimes?

dseevr:
Looks like this is ready to merge!

Vagrantfile
dseevr: Ahh, never mind.  I saw the second .state file in the system-tests dir.

erikh:
LGTM

PR #327, opened 2016-07-07
volplugin, systemtests: Incorporated mount locks

### LINKED ISSUE #295: Option to lock mounts on the same host
Hi @erikh and everyone,
When I create a container with rbd image, I want to lock rbd image to prevent the other container use it. Mean don't allow two containers use same rbd image. 
I've tried use both unlocked: false and unlocked: true, they didn't work, they allow two containers on same host use same rbd image. So, how can I do that ?
Thanks

yuva29:
fixes #295

### DIFF
--- a/systemtests/battery_test.go
+++ b/systemtests/battery_test.go
@@ -21,6 +21,7 @@ func (s *systemtestSuite) TestBatteryMultiMountSameHost(c *C) {
 func (s *systemtestSuite) BatteryMultiMountSameHost(c *C, isUnlocked string) {
 	totalIterations := 5
 	threadCount := 15
+	totalMounts := 5
 
 	for i := 0; i < totalIterations; i++ {
 		syncChan := make(chan struct{}, threadCount)
--- a/systemtests/battery_test.go
+++ b/systemtests/battery_test.go
@@ -3,77 +3,94 @@ package systemtests
 import (
 	"fmt"
 	"math/rand"
+	"strconv"
 	"strings"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
+
 	. "gopkg.in/check.v1"
 
-	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/remotessh"
 )
 
 func (s *systemtestSuite) TestBatteryMultiMountSameHost(c *C) {
-	outerCount := 5
-	count := 15
+	s.BatteryMultiMountSameHost(c, "true")  //unlocked mount
+	s.BatteryMultiMountSameHost(c, "false") //locked mount
+}
+func (s *systemtestSuite) BatteryMultiMountSameHost(c *C, isUnlocked string) {
+	totalIterations := 5
+	threadCount := 15
 
-	for i := 0; i < outerCount; i++ {
-		syncChan := make(chan struct{}, count)
-		volumes := genRandomVolumes(count)
+	for i := 0; i < totalIterations; i++ {
+		syncChan := make(chan struct{}, threadCount)
+		volumes := genRandomVolumes(threadCount)
 
-		for x := 0; x < count; x++ {
-			go func(volName string, x int) {
+		for workerID := 0; workerID < threadCount; workerID++ {
+			go func(volume string, workerID int) {
 				defer func() { syncChan <- struct{}{} }()
-				fqVolName := fqVolume("policy1", volName)
+				fqVolName := fqVolume("policy1", volume)
 
-				c.Assert(s.createVolume("mon0", fqVolName, nil), IsNil)
-				out, err := s.dockerRun("mon0", false, true, fqVolName, "sleep 10m")
-				c.Assert(err, IsNil, Commentf("Output: %s", out))
-				second, err := s.dockerRun("mon0", false, true, fqVolName, "sleep 10m")
-				log.Debug(second, err)
-				c.Assert(err, IsNil, Commentf("Output: %s", second))
+				c.Assert(s.createVolume("mon0", fqVolName, map[string]string{"unlocked": isUnlocked}), IsNil)
+				containerID, err := s.dockerRun("mon0", false, true, fqVolName, "sleep 10m")
+				c.Assert(err, IsNil, Commentf("Output: %s", containerID))
+
+				for x := 0; x < totalMounts; x++ {
+					id, err := s.dockerRun("mon0", false, true, fqVolName, "sleep 10m")
+					if unlocked, _ := strconv.ParseBool(isUnlocked); unlocked {
+						c.Assert(err, IsNil, Commentf("Output: %s", id))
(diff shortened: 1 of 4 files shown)


volplugin/handlers.go
```diff
@@ -260,8 +260,16 @@ func (dc *DaemonConfig) mount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	volName := volConfig.String()
+	mountCount:= dc.increaseMount(volName)
 
-	if dc.increaseMount(volName) > 1 {
+	if !volConfig.Unlocked && mountCount > 1 {
+		if mountCount > 2 { // keeping counter=2 during locked mounts helps docker unmount from removing the actual mount
```
erikh: I don't think we want to do this decrease here. If the mount is requested multiple times in parallel this will inflate the value, but still cause docker to issue an unmount and eventually go below 0 (which deliberately causes a panic). Let's somehow make sure this is tested directly if you feel the need to keep it.

systemtests/battery_test.go
```diff
@@ -13,67 +14,76 @@ import (
...
+		for workerId:= 0; workerId <threadCount; workerId++ {
+			go func(workerId int) {
 				defer func() { syncChan <- struct{}{} }()
-				c.Assert(s.createVolume("mon0", "policy1", fmt.Sprintf("test%02d", x), nil), IsNil)
-				out, err := s.dockerRun("mon0", false, true, fmt.Sprintf("policy1/test%02d", x), "sleep 10m")
-				c.Assert(err, IsNil, Commentf("Output: %s", out))
-				second, err := s.dockerRun("mon0", false, true, fmt.Sprintf("policy1/test%02d", x), "sleep 10m")
-				log.Debug(second, err)
-				c.Assert(err, IsNil, Commentf("Output: %s", second))
+				c.Assert(s.createVolume("mon0", "policy1", fmt.Sprintf("test%02d", workerId), map[string]string{"unlocked":isUnlocked}), IsNil)
```
erikh: Is this working with NFS?

systemtests/volplugin_test.go
```diff
@@ -170,7 +170,7 @@ func (s *systemtestSuite) TestVolpluginRestartMultiMount(c *C) {
 	_, err := s.mon0cmd("sudo truncate -s0 /tmp/volplugin.log")
 	c.Assert(err, IsNil)
 
-	c.Assert(s.createVolume("mon0", "policy1", "test", nil), IsNil)
+	c.Assert(s.createVolume("mon0", "policy1", "test", map[string]string{"unlocked":"true"}), IsNil)
```
erikh: same here re: nfs

erikh:
I'm guessing the passing tests just answered my questions re: NFS. Missed that sorry.

I think we just need a very aggressive test to trigger the mount locking race I think we still have here. If not, we still have an extra test, so this should still be a good use of time. Let's discuss offline how that test might look.

dseevr:
Only comment I have is that you might want to configure your editor to automatically run `gofmt` on your code when you save it.  There's a few spacing issues, etc. but nothing serious.

erikh:
Hmm that makes me think it might be time to add gofmt checks like our golint/vet ones, etc.

yuva29:
That sounds goods to add gofmt along with other go tools.

erikh:
#328 provided for future expansion.

volplugin/handlers.go
yuva29: Good catch. You are right. It is not required and In fact that code path will never be executed as consecutive unmount after mount failure will always keep counter at 1 and never > 2. I'm working on fixing this.

yuva29:
#333

unclejack:
This needs to be rebased on master.

systemtests/volplugin_test.go
```diff
@@ -170,7 +167,7 @@ func (s *systemtestSuite) TestVolpluginRestartMultiMount(c *C) {
 	_, err := s.mon0cmd("sudo truncate -s0 /tmp/volplugin.log")
 	c.Assert(err, IsNil)
 
-	c.Assert(s.createVolume("mon0", "policy1", "test", nil), IsNil)
+	c.Assert(s.createVolume("mon0", "policy1", "test", map[string]string{"unlocked": "true"}), IsNil)
```
erikh: this should not be unlocked.

yuva29: If its not unlocked, multiple mounts won't be possible which will contradict the test case.

erikh: ohhh becuase it's same host. ok, thanks.

yuva29:
This patch will not do graceful unmount due to this issue #333 

Until docker fixes the issue, its safe to use this mount format policy1/test:/mnt:no copy

cc @erikh @dseevr @unclejack

systemtests/battery_test.go
```diff
@@ -2,78 +2,93 @@ package systemtests
 
 import (
 	"fmt"
+	. "gopkg.in/check.v1"
```
erikh: Please put this line, surrounded by whitespace lines, on line 11. This is where https://github.com/bradfitz/goimports would put it by default (we can discuss tooling further offline if you want)

systemtests/battery_test.go
```diff
@@ -2,78 +2,93 @@ package systemtests
...
-		for x := 0; x < count; x++ {
+		for x := 0; x < threadCount; x++ {
 			<-syncChan
 		}
 
 		if cephDriver() {
-			_, err := s.mon0cmd("mount | grep -q rbd")
-			c.Assert(err, IsNil)
+			out, _ := s.mon0cmd("mount | grep -q rbd")
+			c.Assert(out, Equals, "")
```
erikh: this will always succeed: because `-q` never prints anything, just returns an exit status.

Please revert this section to the left hand side, it was already correct.

volplugin/handlers.go
```diff
@@ -261,7 +261,14 @@ func (dc *DaemonConfig) mount(w http.ResponseWriter, r *http.Request) {
 
 	volName := volConfig.String()
 
-	if dc.increaseMount(volName) > 1 {
+	// XXX docker issues unmount request after every mount failure so, this evens out
+	//     decreaseMount() in unmount
+	mountCount := dc.increaseMount(volName)
+	if !volConfig.Unlocked && mountCount > 1 {
```
erikh: let's reorganize this a bit:

```go
if mountCount > 1 {
  if !volConfig.Unlocked {
    first block
    return
  }

  second block
  return
}
```

Pretty sure this behaves the same and is a little easier on the eyes IMO

yuva29: yeah they are same. I'll change that

dseevr:
@unclejack it's rebased now

@yuva29 is this ready for final review?

yuva29:
@dseevr Yeah.

erikh:
LGTM

erikh:
can everyone review this today? This is the last mile for v0.2.

dseevr:
LGTM

PR #330, opened 2016-07-07
Improve test performance

erikh:
The volume CRUD primitives were lousy. I moved on @mapuri's randomized
containers patches and did the same for volumes. Because of a bug in docker
(see below) we cannot create the same volume twice in a short period of time
without hitting a cache bug:

docker/docker#24205 for more information.

This fix implements a cleaner API for managing volumes, and converts the whole
suite to use those features, including the randomized volume names.

### DIFF
--- a/systemtests/util_test.go
+++ b/systemtests/util_test.go
@@ -22,6 +22,44 @@ var startedContainers struct {
 	names map[string]struct{}
 }
 
+var (
+	volumeListMutex = sync.Mutex{}
+	volumeList      = map[string]struct{}{}
+)
+
+func volumeParts(volume string) (string, string) {
+	parts := strings.SplitN(volume, "/", 2)
+	return parts[0], parts[1] // panic, schmanick
+}
+
+func fqVolume(policy, volume string) string {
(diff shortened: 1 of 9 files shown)


erikh:
This patch improves test performance by over 200%

systemtests/util_test.go
```diff
@@ -22,6 +22,44 @@ var startedContainers struct {
...
+func volumeParts(volume string) (string, string) {
+	parts := strings.SplitN(volume, "/", 2)
+	return parts[0], parts[1] // panic, schmanick
+}
+
+func fqVolume(policy, volume string) string {
+	return strings.Join([]string{policy, volume}, "/")
+}
+
+func genRandomVolume() string {
```
unclejack: Could this run into collisions because genRandomString doesn't avoid collisions?

erikh: the next 10 lines of code are collision detection and retry.

unclejack: Ok, got it.

unclejack:
LGTM

dseevr:
LGTM

yuva29:
LGTM

PR #331, opened 2016-07-09
Gofmt

erikh:
This implements gofmt checks for the source when building (and thusly testing) the source.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,12 @@ provision:
 ssh:
 	vagrant ssh mon0
 
+gofmt-host:
+	@build/scripts/gofmt.sh
+
+gofmt:
+	vagrant ssh mon0 -c "sudo -i sh -c 'cd $(GUESTGOPATH); http_proxy=${http_proxy} https_proxy=${https_proxy} make gofmt-host'"
+
 golint-host:
 	[ -n "`which golint`" ] || go get github.com/golang/lint/golint
 	set -e; for i in $$(go list ./... | grep -v vendor); do golint $$i; done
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -44,6 +44,12 @@ provision:
 ssh:
 	vagrant ssh mon0
 
+gofmt-host:
+	@build/scripts/gofmt.sh
+
+gofmt:
```
yuva29: Why do we need gofmt inside VM ?

build/scripts/gofmt.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh	
+
+out=$(gofmt -l -s **/*.go | grep -v vendor)
```
yuva29: If we intend to run gofmt on every .go file, this pattern ignores go files inside subfolder (e.g apiserver/apiserver/cli.go)

Makefile
erikh: we need to be consistent building and checking the source, so we do this inside the VM

build/scripts/gofmt.sh
erikh: I'll fix this in a bit; bash vs. zsh issue.

build/scripts/gofmt.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh	
+
+out=$(find . -name '*.go' -type f | xargs gofmt -l -s | grep -v vendor)
```
dseevr: Should we use `egrep -v '^vendor/'` just to be extra cautious?

I think the `grep` and `xargs` are out of order, too.

`out=$(find . -name '*.go' -type f | egrep -v '^vendor/' | xargs gofmt -l -s)`

erikh: it doesn't matter, the result will be grepped out either way.

dseevr: Oh, I was thinking we didn't want to `gofmt` anything under /vendor

erikh: it doesn't write the files, just checks them.

erikh: aka, after the grep it's a noop.

dseevr: Ah, sorry, I was mistaking -s for -w.  My bad.

dseevr:
LGTM

PR #336, opened 2016-07-12
apiserver,config: refactor Global configuration management (see below)

### LINKED ISSUE #334: Global parameters are a pain to edit
Right now the fixup of the global parameters (converting from time.Duration to an integer with a fixed base like time.Minute) does not "un-fix" on the way out, causing large `uint64`s to be sprayed in the json, which then is multiplied again if an astute reader is not watching.

This is legacy from when volplugin would contact the volmaster for these values and needed an unfiltered version.

volcli does the division right now, but it really should be done in the API server.

We should divide these parameters on the API endpoint instead.

/cc @melanietom

erikh:
Globals were pretty ugly in how they calculated parameters. They were stateful
and it was hard to manage between the different spots in the code.

I refactored it to use a stateless design with three new functions to set
parameters in a functional way. This way instead of passing around mutable
state, code can just ask for the type of parameters it needs.

Fixes #334 

/cc @melanietom 

### DIFF
--- a/apiserver/daemon.go
+++ b/apiserver/daemon.go
@@ -493,7 +493,7 @@ func (d *DaemonConfig) handleCopy(w http.ResponseWriter, r *http.Request) {
 }
 
 func (d *DaemonConfig) handleGlobal(w http.ResponseWriter, r *http.Request) {
-	content, err := json.Marshal(d.Global)
+	content, err := json.Marshal(d.Global.Published())
 	if err != nil {
 		api.RESTHTTPError(w, errors.MarshalGlobal.Combine(err))
 		return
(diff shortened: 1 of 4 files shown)


dseevr:
LGTM

yuva29:
Globals are displayed in standard formats(second, minute) to users but internally they are maintained as duration. Is it what we are trying to achieve ?

erikh:
Correct; the `time.Duration` is needed internally to simplify the code, and the end-users probably don't want to specify their values in nanoseconds (which is what time.Duration really is)

yuva29:
Got it. So, what we specify in globals.json is nanoseconds?

erikh:
In the JSON you specify it will be in second/minute representation. Internally the nanosecond version is used.

yuva29:
LGTM

PR #342, opened 2016-07-14
Makefile: turn off release builds

erikh:
The release builds, while nice when we were pre-versioned are broken; they still tag at v0.0.0. Additionally, they add a lot of noise to the releases page which drown the actual point releases' notes.

I think soon our release tooling will need to be revisited, but this is a first step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,6 +137,9 @@ system-test: run
 system-test-big:
 	BIG=1 make system-test
 
+vendor-ansible:
+	git subtree pull --prefix ansible https://github.com/contiv/ansible HEAD --squash
+
 reflex:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -137,6 +137,9 @@ system-test: run
 system-test-big:
 	BIG=1 make system-test
 
+vendor-ansible:
+	git subtree pull --prefix ansible https://github.com/contiv/ansible HEAD --squash
+
```
erikh: oh. I moved this up to be closer to the other build/dev related tasks.

dseevr:
LGTM

erikh:
I don't want this to happen again tonight so I don't have to run the cleanup again. I'm merging this now.

PR #343, opened 2016-07-14
README.md,config,systemtests,volplugin: remove IOPS controls.

erikh:
This removes IOPS controls; we could not guarantee these values for either Ceph or NFS. The BPS rate limiting for Ceph is still applied, and as soon as I can figure out what I need to do with `tc` to handle the NFS side, I will.

Opinions on the change are encouraged. /cc @mapuri @jainvipin

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ release.
 * Manage many kinds of filesystems, including providing mkfs commands.
 * Snapshot frequency and pruning. Also copy snapshots to new volumes!
 * Ephemeral (removed on container teardown) volumes
-* IOPS limiting (via blkio cgroup)
+* BPS limiting (via blkio cgroup)
 
 volplugin is still alpha at the time of this writing; features and the API may
 be extremely volatile and it is not suggested that you use this in production.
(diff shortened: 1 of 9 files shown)


erikh:
I will also note that glusterfs has a ton of quota management tools and we will be implementing those, trying to find ceph/nfs analogues as soon as the time is available.

dseevr:
Code changes LGTM

yuva29:
Please explain me why were are not supporting IOPS ?

erikh:
Sure; basically, iops are a per-system weight and not a global weight. In a
shared storage environment, you cannot reliably predict the i/o load each
server will send to the cluster, and you cannot reliably predict the i/o
load on each individual server serving the storage, since weights are
per-system.

I have some ideas on how to manage this but ceph doesn't have any support
in krbd to provide the data I need. We can speak of this strategy privately
if you'd like to know more.

erikh:
Note that we may be able to support more advanced stuff once glusterfs
support comes, it's just not here yet to play with.

yuva29:
Sure. Definitely would like to know more! 

Changes LGTM.

erikh:
RFC expired, please raise concerns here if you still have them @jainvipin @mapuri

Issue #332, opened 2016-07-11
vagrant reload breaks virtualbox


erikh:
This causes the disk errors we typically see when vbox is throttled. Perhaps we should solve this by checking for the file existence before both removing and creating it.

This will need changes to both Makefile and Vagrantfile

dseevr:
I made the change so that the disks are only created/attached if they don't already exist, but I'm encounter a new problem:  the shared directories are failing to mount during `vagrant reload`

```

**Comment by dseevr** [comment]
I made the change so that the disks are only created/attached if they don't already exist, but I'm encountering a new problem:  the shared directories are failing to mount during `vagrant reload`

```
==> mon0: Configuring and enabling network interfaces...
==> mon0: Mounting shared folders...
    mon0: /tmp/bin => /home/bilrobin/src/github.com/contiv/volplugin/bin
Failed to mount folders in Linux guest. This is usually because
the "vboxsf" file system is not available. Please verify that
the guest additions are properly installed in the guest and
can work properly. The command attempted was:

mount -t vboxsf -o uid=`id -u vagrant`,gid=`getent group vagrant | cut -d: -f3` tmp_bin /tmp/bin
mount -t vboxsf -o uid=`id -u vagrant`,gid=`id -g vagrant` tmp_bin /tmp/bin

The error output from the last command was:

/sbin/mount.vboxsf: mounting failed with the error: No such device
```

One thing I noticed is that the vbox drivers are not loaded.  This is what they look like on a freshly initialized environment (via `make restart`):

```
[vagrant@mon0 ~]$ lsmod | grep vbox
vboxsf                 39741  3
vboxguest             297222  2 vboxsf
vboxvideo              12658  0
drm                   349210  2 vboxvideo
```

On a reloaded environment, that command has no results.

dseevr:
It turns out this new issue is related to the fact that part of our Ansible setup updates all the packages on the system which *may* include the kernel.  If a new kernel is installed, the vbox guest drivers are not compiled for it and will not be loaded at the next boot (i.e., the `vagrant reload` this PR is about).

dseevr:
Manually fixing this and `vagrant reload`ing results in the guests coming up without Docker running.  This is causing etcd not to start which causes tests to fail.  Looking into this now.

dseevr:
With #346 and #347 merged in, the new `make reload` target successfully reboots the VMs which technically completes this issue.

[...]

PR #346, opened 2016-07-14
Vagrantfile: skip creating vbox disk image if it already exists

dseevr:
This was one of the blockers for `vagrant reload` working successfully. https://github.com/contiv/volplugin/issues/332

The other was the kernel being upgraded by Ansible resulting in the vbox drivers not being loaded after reboot, which broke shared folders.  This was addressed by @mapuri here: https://github.com/contiv/ansible/pull/245

cc @erikh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,9 @@ update:
 
 restart: stop update start
 
+reload:
+	vagrant reload
+
 provision:
 	vagrant provision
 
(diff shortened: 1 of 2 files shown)


dseevr:
This has been tested, and works, but it depends on the redhat_tasks.yml change in https://github.com/contiv/volplugin/pull/347

erikh:
LGTM

erikh:
Gonna merge this one as it's a simple change and we don't all need to review it.

PR #352, opened 2016-07-15
Makefile: add --provision flag to 'reload' target to ensure services are started after VM reboot

dseevr:
This flag should have been added in #346.

See:  https://github.com/contiv/volplugin/issues/332#issuecomment-233085163

Thanks, @mapuri!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ update:
 restart: stop update start
 
 reload:
-	vagrant reload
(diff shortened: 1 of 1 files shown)


dseevr:
Just gonna merge this because it's a trivial change

erikh:
merged because it's a build env change and won't be tested anyways.

PR #355, opened 2016-07-18
api/docker,volplugin: new api/docker package for storing docker api top-level handlers.

erikh:
Mostly removing and moving around a bit of code. No real logic changes.


dseevr:
LGTM

PR #356, opened 2016-07-18
Makefile,build/scripts: adjust systemtests to have independent target…

erikh:
…s to be later used by jenkins

@mapuri


dseevr:
LGTM

mapuri:
LGTM

mapuri:
@erikh , do you want to get rid of the `make ci` target as well? I don't think we would need it anymore with parallel builds

erikh:
@mapuri I think I will do that in a separate patch so I don't have to run tests for this one again.

PR #357, opened 2016-07-18
More API porting

erikh:
This moves some more stuff into `api/` from `volplugin/`, specifically the mount counting and mapping code.


dseevr:
LGTM

PR #360, opened 2016-07-19
build/scripts: fix systemtest logic

mapuri:
- remove the loop otherwise the tests run twice adding to test time
- also remove setting GOPATH for the CI environment in scripts, it is better to control it from higher level jenkins job or make target

I found some of these while trying out parallel tests.

/cc @erikh


erikh:
ah! thanks. I'll wait for tests to run then merge.

erikh:
sorry about that!

PR #363, opened 2016-07-25
Makefile: clean up global state associated with the vagrant setup

mapuri:
This will help cleanup state on CI machines, once the job has completed. It will be useful otherwise too I think.

/cc @dseevr @erikh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ stop:
 	make clean
 
 clean:
+	([ -n "$$(cat subnet_assignment.state)" ] && rm -rf /tmp/volplugin_vagrant_subnets/`cat subnet_assignment.state`) || :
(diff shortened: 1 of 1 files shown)


dseevr:
LGTM

Makefile
```diff
@@ -23,6 +23,7 @@ stop:
 	make clean
 
 clean:
+	[ "$$(cat subnet_assignment.state)" = "" ] || rm -rf /tmp/volplugin_vagrant_subnets/`cat subnet_assignment.state`  || true
```
erikh: * Use `-n` negated to test the empty string
* Instead of using `true`, use `:`.

mapuri:
@erikh I took care of the comments, PTAL when you get a chance.

PR #366, opened 2016-07-26
JSON Validation

### LINKED ISSUE #237: Compatibility checks for backends
#231 provides user a way to specify the different backends to perform volplugin operations viz. CRUD, mount and snapshots.

This issue tracks the discussion about handling multiple backends at policy definition time (https://github.com/contiv/volplugin/pull/231#issuecomment-211478222)

Following are a few thoughts on possible ways of handling/providing mix of backends:

### Validate user input

This extends the way of policy definition as as introduced by #231. In this mode we let user specify the different backends and then we shall validate the user input against a compatibility matrix of supported backends.
**Pros**:
- let's user choose the backends of their choice as long as they are compatible 
- can be very useful to mix and match compatibile driver, assuming multiple such backends exist 

**Cons**:
- user needs to explicitly specify each backend to be able to avail behavior. For instance, user will need to explicitly specify snapshot driver to be able to take advantage of a backend's snapshot capability
- a complex compatibility matrix can exist, assuming we have multiple backends that can be compatible with each other.

### Derive compatible backends
In this mode we can accept just one backend name from the user at policy definition time i.e. keep the behavior as it was before #231. And then we initialize the different backends based on the user specified backend. For instance, when user specifies `ceph` then we shall initialize all 3 backends to be ceph, but when user specifies `nfs` we may only initialize `mount` backend and leave rest as empty

**Pros**
- simple compatibility checks
- frees user of having to understand compatibility of different backends. They still need to understand capabilities of each backend, which is simpler

**Cons**
- lacks the rick mix and match of different compatible backends

yuva29:
Fixes #237
1. This simplifies “Policy” and “Volume” validations using JSON schema
2. Refactor driver creation code (NewDriver)
3. More rules can be added to the schema. Right now, the validation rules are defined based on earlier validation.


storage/backend/backends.go
```diff
@@ -40,20 +44,41 @@ func NewMountDriver(backend, mountpath string) (storage.MountDriver, error) {
...
+			} else if err := mount.Validate(do); err != nil {
+				return err
+			}
+		} else if driverType == Snapshot {
+			if snapshot, err := NewSnapshotDriver(backend); err != nil {
+				return err
+			} else if err := snapshot.Validate(do); err != nil {
+				return err
+			}
+		}
```
dseevr: I think there should be an else here which returns an error, especially because this function is exported.

config/volume.go
```diff
@@ -343,35 +347,34 @@ func (c *Client) WatchSnapshotSignal(activity chan *watch.Watch) {
...
-func (ro *RuntimeOptions) Validate() error {
-	if ro.UseSnapshots && (ro.Snapshot.Frequency == "" || ro.Snapshot.Keep == 0) {
-		return errored.Errorf("Snapshots are configured but cannot be used due to blank settings")
-	}
-
-	return nil
-}
-
-// Validate validates a volume configuration, returning error on any issue.
+// Validate validates a volume configuration, returning error on any
```
dseevr: Looks like this line was accidentally edited.

storage/backend/backends.go
```diff
@@ -26,10 +33,7 @@ var SnapshotDrivers = map[string]func() (storage.SnapshotDriver, error){
 // NewMountDriver instantiates and return a mount driver instance of the
 // specified type
 func NewMountDriver(backend, mountpath string) (storage.MountDriver, error) {
-	f, ok := MountDrivers[backend]
-	if !ok {
-		return nil, errored.Errorf("invalid mount driver backend: %q", backend)
-	}
+	f, _ := MountDrivers[backend]
```
dseevr: You can omit the second value from the assignment entirely if you're not checking for existence when plucking a value from a map.

```go
f := MountDrivers[backend]
```

If it doesn't exist, you just get the default uninitialized state for the type:  https://play.golang.org/p/HXax59236C

Why are we switching from checking to see if the backend exists to assuming it exists?  I'm not familiar with this code yet, so I'm just curious.

storage/backend/backends.go
yuva29: Yep. You're right. I'll handle that case.

storage/backend/backends.go
yuva29: Cool 👍  Validation always ensures that 'backend' will be one of the allowed drivers.

https://github.com/yuva29/volplugin/blob/674460acf3628f7eb45ffdd6fbdc7e260fdc141a/config/schema.go#L34-L36

String empty checks are handled in NewDriver.

But now, I think it is better to check the existence of them as its an exported function.

config/volume.go
yuva29: These checks are moved to JSON schema. 

Schema here: https://github.com/yuva29/volplugin/blob/674460acf3628f7eb45ffdd6fbdc7e260fdc141a/config/schema.go#L5-L23 

Validation here: https://github.com/yuva29/volplugin/blob/674460acf3628f7eb45ffdd6fbdc7e260fdc141a/config/validation.go#L23-L33

dseevr: I was referring to the deletion of " issue."  The sentence is incomplete in the latest commit.  It looks like the last word was accidentally deleted.

yuva29: Sorry about that. I'll handle this.

config/policy.go
```diff
@@ -123,20 +130,20 @@ func (cfg *Policy) Validate() error {
...
-		return errored.Errorf("Mount backend cannot be empty for policy %v", cfg)
+	if err := ValidatePolicy(cfg); err != nil {
+		return errors.ErrJSONValidation.Combine(err)
 	}
 
-	size, err := cfg.CreateOptions.ActualSize()
-	if cfg.Backends.CRUD != "" && (size == 0 || err != nil) {
-		return errored.Errorf("Size set to zero for non-empty CRUD backend %v", cfg.Backends.CRUD).Combine(err)
+	if cfg.Backends.Mount == "" { // backend should be defined and its validated
+		cfg.Backends = defaultDrivers[cfg.Backend]
```
erikh: reading from maps has a secondary form:

```go
var, ok := myMap[myKey]
if ok {
  // key exists
}
```

This is really important here because we're not using pointers, and therefore this will always succeed.

If we don't have a default driver, perhaps we should throw an error? I'm not sure what the best option is here, ideas welcome.

config/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/contiv/errored"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+
+	gojson "github.com/xeipuuv/gojsonschema"
+)
+
+// Combines array of errors into a single error
+func combineErrors(resultErrors []gojson.ResultError) error {
```
erikh: use `errored.Combine` here please and the `errors` package to prefix it with a canned error.

erikh: I'm not sure this function is necessary to exist after this change.

config/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	var errors []string
+	for _, err := range resultErrors {
+		errors = append(errors, fmt.Sprintf("%s\n", err))
+	}
+	return errored.New(strings.Join(errors, "\n"))
+}
+
+// ValidateRuntime validates the given runtime against its defined schema
+func ValidateRuntime(cfg *RuntimeOptions) error {
```
erikh: Please call this Validate, and make the receiver `*RuntimeOptions`.

config/validation.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if result, _ := gojson.Validate(schema, doc); !result.Valid() {
+		return combineErrors(result.Errors())
+	}
+	log.Infof("Runtime config validation passed")
+	return nil
+}
+
+// ValidatePolicy validates the given policy against its defined schema
+func ValidatePolicy(cfg *Policy) error {
```
erikh: Same here.

config/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	schema := gojson.NewStringLoader(PolicySchema)
+	doc := gojson.NewGoLoader(cfg)
+
+	if result, _ := gojson.Validate(schema, doc); !result.Valid() {
```
erikh: is there an error returned here? If so, we need to check it.

config/validation.go
```diff
@@ -0,0 +1,66 @@
...
+	if err := ValidateRuntime(&cfg.RuntimeOptions); err != nil {
+		return err
+	}
+
+	log.Infof("Policy validation passed")
+	return nil
+}
+
+// ValidateVolume validates the given volume against its defined schema
+func ValidateVolume(cfg *Volume) error {
```
erikh: Same here re: receiver.

config/volume.go
```diff
@@ -343,35 +347,34 @@ func (c *Client) WatchSnapshotSignal(activity chan *watch.Watch) {
...
 
-	if cfg.PolicyName == "" {
-		return errored.Errorf("Policy name was omitted for volume %v", cfg)
-	}
+	// We use a few dummy variables to ensure that global configuration is
+	// not needed in the storage drivers, that the validation does not fail
+	// because of it.
+	var mountPath string
+	for driverType, backendName := range map[string]string{backend.Mount: cfg.Backends.Mount, backend.CRUD: cfg.Backends.CRUD, backend.Snapshot: cfg.Backends.Snapshot} {
+		defer func() { mountPath = "" }()
```
erikh: don't set variables in defer; also I don't think this will fire on time for it to be useful. You're also doing it in a loop which will allocate a function for each entry, which I don't believe is necessary.

config/volume.go
```diff
@@ -343,35 +347,34 @@ func (c *Client) WatchSnapshotSignal(activity chan *watch.Watch) {
...
 
-	if err := cfg.validateBackends(); err != nil {
-		return err
+		if driverType == backend.Mount {
+			mountPath = "/mnt"
```
erikh: there are constants for this in storage; make them public if necessary and use them instead of static strings.

storage/backend/backends.go
```diff
@@ -57,3 +64,31 @@ func NewSnapshotDriver(backend string) (storage.SnapshotDriver, error) {
 
 	return f()
 }
+
+// NewDriver creates a driver based on driverType
+func NewDriver(backend string, driverType string, mountPath string, do storage.DriverOptions) error {
+	if backend != "" {
+		if driverType == CRUD {
+			if curd, err := NewCRUDDriver(backend); err != nil {
```
erikh: misspelled `crud`

storage/backend/backends.go
erikh: no need for the else, the return below will suffice, and using an else here will cause a go vet error.

dseevr: I was thinking logic-wise... is it correct for there not to be an else case here?  Seems like we would want to raise an error if someone specifies a non-CRUD/-mount/-validate value in driverType rather than falling through to the final return.

erikh:
Final thoughts on first pass:

* Get familiar with our local `errors` library and contiv/errored. You are not appropriately converting the types so we can get meaningful tracing and error messages.
* Check your errors. You just missed a few things I think.
* it may be the right time to move our internal struct members to use pointers instead of copies. A number of issues I mentioned will be solved by this, but please investigate this and see if it would make things easier for you. Your discretion.

storage/backend/backends.go
erikh: let's use a switch statement and default here, this is noisy.

erikh: it will also clarify the logic, I think.

config/validation.go
yuva29: there is already a validate in config/policy.go(func (*Policy) Validate()) . If I define it here, it will be a duplicate.

erikh: Combine the methods?

yuva29: I can do it. But I wanted the schema validation to be in separate function, just not to make Validate functions noisy. Please input your thoughts.

config/policy.go
yuva29: Yes. You're right. I did not check existence because cfg.Backend will always contain one of the available drivers (ceph & nfs as of now, other values will not pass the validation ). But I'll check for existence as a safety measure.

config/validation.go
erikh: That's fair; Let's do this then. Call these functions `ValidateJSON' or
something similar; call into them that way instead with the receivers
configured.

config/validation.go
yuva29: gojsonschema package returns array of errors([]gojson.ResultError) that violated the validation rule. Like below:
backends.mount: backends.mount must be one of the following: "ceph", "nfs"
backends.mount: String length must be greater than or equal to 1

I'm combining all the errors into a single string and creating a error using errored.New()

I can use ```(*errored.Error).Combine``` but there will be stack trace for each one of them (Is it necessary? ) in debug mode. Also, list of errors can be huge based on  the JSON.

Please input your thoughts.

erikh: SGTM, didn't realize that was the goal here.

erikh:
LGTM. As mentioned on slack; tests will be added for this post-hoc.

@dseevr PTAL

dseevr:
I don't have any further input.  LGTM

erikh:
@yuva29 is this ready? I noticed you pushed recently.

yuva29:
I fixed gofmt issues with one of the file. Also, I noticed some instability with ```systemtestSuite.TestBatteryMultiMountSameHost``` so I'm running the tests again to confirm instability. I would like to wait for CI completion.

yuva29:
@erikh This can be merged now.

PR #367, opened 2016-07-26
apiserver,config,errors,volcli: add volcli commands to get, list, and watch policy revision changes

### LINKED ISSUE #269: Policy audit-logging / journal
Policies cover volumes over time and may be modified, so the policy may drift over time (and the volumes themselves) so far that the volumes no longer resemble the policies they were created under

It would be nice to cross-reference these somehow, but for now a timestamped journal would be ideal. This would also be very useful for audit logging later.

I'm not sure if this information should live in etcd or be dumped elsewhere. I think that merits a conversation.

dseevr:
- Get all revision numbers: `volcli policy history list $name`
- Retrieve the policy text for a specific revision: `volcli policy history get $name $revision`
- Watch for new policies being uploaded: `volcli policy watch`

Fixes #269

### DIFF
--- a/config/archive.go
+++ b/config/archive.go
@@ -1,0 +1,89 @@
+package config
+
+import (
+	"fmt"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/contiv/volplugin/errors"
+	"github.com/contiv/volplugin/watch"
+	"github.com/coreos/etcd/client"
+	"golang.org/x/net/context"
+)
+
+func (c *Client) policyArchive(name string) string {
+	return c.prefixed(rootPolicyArchive, name)
+}
+
+func (c *Client) policyArchiveEntry(name string, timestamp string) string {
+	return c.prefixed(rootPolicyArchive, name, timestamp)
+}
+
+// CreatePolicyRevision creates an revision entry in a policy's history.
+func (c *Client) CreatePolicyRevision(name string, policy string) error {
+	timestamp := fmt.Sprint(time.Now().Unix())
+	key := c.policyArchiveEntry(name, timestamp)
+
+	_, err := c.etcdClient.Set(context.Background(), key, policy, nil)
+	if err != nil {
+		return errors.EtcdToErrored(err)
+	}
+
+	return nil
+}
+
+// ListPolicyRevisions returns a list of all the revisions for a given policy.
+func (c *Client) ListPolicyRevisions(name string) ([]string, error) {
+	keyspace := c.policyArchive(name)
+
+	resp, err := c.etcdClient.Get(context.Background(), keyspace, &client.GetOptions{Sort: true, Recursive: false, Quorum: true})
+	if err != nil {
+		return nil, errors.EtcdToErrored(err)
+	}
+
+	ret := []string{}
+
+	for _, node := range resp.Node.Nodes {
+		_, revision := path.Split(node.Key)
+
+		ret = append(ret, revision)
+	}
+
+	return ret, nil
+}
+
+// GetPolicyRevision returns a single revision for a given policy.
+func (c *Client) GetPolicyRevision(name, revision string) (string, error) {
+	keyspace := c.policyArchiveEntry(name, revision)
+
+	resp, err := c.etcdClient.Get(context.Background(), keyspace, &client.GetOptions{Sort: false, Recursive: false, Quorum: true})
+	if err != nil {
+		return "", errors.EtcdToErrored(err)
+	}
+
+	return resp.Node.Value, nil
+}
+
+// WatchForPolicyChanges creates a watch which returns the policy name and revision
+// whenever a new policy is uploaded.
+func (c *Client) WatchForPolicyChanges(activity chan *watch.Watch) {
+	keyspace := c.prefixed(rootPolicyArchive)
+
+	w := watch.NewWatcher(activity, keyspace, func(resp *client.Response, w *watch.Watcher) {
+		// skip anything not happening _below_ our watch.
+		// we don't care about updates happening to the root of our watch here.
+		if !strings.HasPrefix(resp.Node.Key, keyspace+"/") {
+			return
+		}
+
+		s := strings.TrimPrefix(resp.Node.Key, keyspace+"/")
+
+		name, revision := path.Split(s)
+		name = name[:len(name)-1] // remove trailing slash
+
+		w.Channel <- &watch.Watch{Key: resp.Node.Key, Config: []string{name, revision}}
+	})
+
+	watch.Create(w)
+}
--- a/config/archive_test.go
+++ b/config/archive_test.go
@@ -1,0 +1,70 @@
+package config
+
+import (
+	"github.com/contiv/volplugin/watch"
+	. "gopkg.in/check.v1"
+)
+
+const (
+	testPolicyName = "foobar"
+)
+
+func revisionCount(s *configSuite, c *C) int {
+	revisions, err := s.tlc.ListPolicyRevisions(testPolicyName)
+	if err != nil {
+		// this is a "does not exist" because the policy hasn't been uploaded yet
+		return 0
+	}
+
+	return len(revisions)
+}
+
+func createTestPolicy(s *configSuite, c *C) {
+	err := s.tlc.CreatePolicyRevision(testPolicyName, "{}")
+	c.Assert(err, IsNil)
+}
+
+func (s *configSuite) TestCreatePolicyRevision(c *C) {
+	c.Assert(revisionCount(s, c), Equals, 0)
+	createTestPolicy(s, c)
+	c.Assert(revisionCount(s, c), Equals, 1)
+}
+
+// the logic for testing Get and List only differs in that testing Get verifies
+// that the policy text is what was uploaded, so the two tests are combined here.
+func (s *configSuite) TestGetAndListPolicyRevisions(c *C) {
+	policyText := `{"foo":"bar"}`
+
+	c.Assert(revisionCount(s, c), Equals, 0)
+
+	err := s.tlc.CreatePolicyRevision(testPolicyName, policyText)
+	c.Assert(err, IsNil)
+
+	revisions, err := s.tlc.ListPolicyRevisions(testPolicyName)
+	c.Assert(err, IsNil)
+	c.Assert(len(revisions), Equals, 1)
+
+	resp, err := s.tlc.GetPolicyRevision(testPolicyName, revisions[0])
+	c.Assert(err, IsNil)
+	c.Assert(resp, Equals, policyText)
+}
+
+func (s *configSuite) TestWatchForPolicyChanges(c *C) {
+	activity := make(chan *watch.Watch)
+	s.tlc.WatchForPolicyChanges(activity)
+
+	createTestPolicy(s, c)
+
+	w := <-activity
+
+	watch.Stop(s.tlc.prefixed(rootPolicyArchive))
+
+	changeset := w.Config.([]string)
+	c.Assert(len(changeset), Equals, 2)
+
+	name := changeset[0]
+	revision := changeset[1]
+
+	_, err := s.tlc.GetPolicyRevision(name, revision)
+	c.Assert(err, IsNil)
+}
--- a/volcli/volcli.go
+++ b/volcli/volcli.go
@@ -18,6 +18,7 @@ import (
 	"github.com/contiv/errored"
 	"github.com/contiv/volplugin/config"
 	"github.com/contiv/volplugin/lock"
+	"github.com/contiv/volplugin/watch"
 	"github.com/kr/pty"
 )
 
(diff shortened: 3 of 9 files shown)


erikh:
If you put 'Fixes #ticket' in your description/commit message, it will automatically close the ticket when this is merged.

config/archive.go
```diff
@@ -0,0 +1,86 @@
...
+func (c *Client) policyArchiveEntry(name string, timestamp string) string {
+	return c.prefixed(rootPolicyArchive, name, timestamp)
+}
+
+// CreatePolicyRevision creates an revision entry in a policy's history.
+func (c *Client) CreatePolicyRevision(name string, policy string) error {
+	timestamp := fmt.Sprint(time.Now().Unix())
+	key := c.policyArchiveEntry(name, timestamp)
+
+	_, err := c.etcdClient.Set(context.Background(), key, policy, &client.SetOptions{PrevExist: client.PrevIgnore})
```
erikh: is there a reason we're not using `PrevExist` here instead of `PrevIgnore`?

config/archive.go
```diff
@@ -0,0 +1,86 @@
...
+		&client.GetOptions{Sort: true, Recursive: false, Quorum: true},
+	)
+	if err != nil {
+		return nil, errors.EtcdToErrored(err)
+	}
+
+	ret := []string{}
+
+	for _, node := range resp.Node.Nodes {
+		parts := strings.Split(node.Key, "/")
```
erikh: `path.Split` is a little safer for this operation.

config/archive.go
```diff
@@ -0,0 +1,86 @@
...
+// WatchForPolicyChanges creates a watch which prints the policy name and revision number
+// (timestamp) to stdout whenever a policy is uploaded.
+func (c *Client) WatchForPolicyChanges(watchChan chan *watch.Watch) {
+	w := watch.NewWatcher(watchChan, c.prefixed(rootPolicyArchive), func(resp *client.Response, w *watch.Watcher) {
+		parts := strings.Split(resp.Node.Key, "/")
+
+		policy := parts[0]
+		timestamp := parts[1]
+
+		fmt.Printf("%s\t%s\n", policy, timestamp)
```
erikh: get rid of this line please, or use `log.Debugf` with some flavor text if you want to keep it.

config/archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/contiv/volplugin/watch"
+	. "gopkg.in/check.v1"
+)
+
+const (
+	testPolicyName = "foobar"
+)
+
+func revisionCount(s *configSuite) int {
+	revisions, _ := s.tlc.ListPolicyRevisions(testPolicyName)
```
erikh: check and log or return errors here.

config/archive.go
dseevr: This is where the function writes to stdout (the purpose of the function).  It's not logging/debugging output

config/archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+	createTestPolicy(s, c)
+	c.Assert(revisionCount(s), Equals, 1)
+}
+
+func (s *configSuite) TestListPolicyRevisions(c *C) {
+	c.Assert(revisionCount(s), Equals, 0)
+
+	// edge case: the second may roll over between saving the current timestamp
+	//            and the HTTP request to upload the policy being handled
+	t := int(time.Now().Unix())
```
erikh: this seems really fragile. Can we find a way to be more deterministic on what timestamp is selected? how about if we just ensure that `revisions[0]` is non-empty?

config/archive_test.go
```diff
@@ -0,0 +1,120 @@
...
+	c.Assert(err, IsNil)
+
+	// edge case: the second may roll over between saving the current timestamp
+	//            and the HTTP request to upload the policy being handled
+	t := int(time.Now().Unix())
+	t2 := t + 1
+
+	// run the stdout manipulation code in a func so we can guarantee a defer to
+	// reset everything runs immediately afterwards even in a panic() situation
+	func() {
```
erikh: this `func` is superfluous... not quite sure what you're trying to accomplish here. Do you just need a new scope?

dseevr: Check out the comment above it.  It's to ensure a panic() will result in stdout being reset back to the default file handle *immediately*  so the rest of the test run isn't eaten.

Doing the reassignment back to the default in a function-level defer would eat any further assertions before the stack is unwound.

volcli/volcli.go
```diff
@@ -286,6 +287,115 @@ func policyList(ctx *cli.Context) (bool, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	c := make(chan *watch.Watch)
+
+	cfg.WatchForPolicyChanges(c)
+
+	for range c {
+	}
```
erikh: were you planning to output the data here?

erikh: I would like that.

config/archive_test.go
erikh: Is `recover` not going to work here?

https://blog.golang.org/defer-panic-and-recover

erikh: Let's find another solution regardless please. I really dislike this code. No reflection on you, to be clear.

erikh:
Please add your root path constants here:

https://github.com/contiv/volplugin/blob/master/config/config.go#L20-L27

This ensures they are safely populated (doesn't matter on write but does matter on read). The constructor below will take care of actually populating the data.

config/archive_test.go
dseevr: The problem is that I need to reset the stdout handle whether anything panics in between taking it and resetting it normally or not.

I guess a function-level defer/recover() could reset it.  It'd be possible for the reset code to be called twice (take stdout, run code, reset it, something else panics, it gets reset again), but that doesn't seem to be a big deal, and it'd handle bother cases.

What do you think about that?

erikh:
Code looks very nice, I'm impressed that you captured a lot of the idioms that were already in the code. Doesn't look like much to do to get this merged.

dseevr:
RE: code path constants

I did do that:  https://github.com/contiv/volplugin/pull/367/commits/4200a64bd5181b9d617c2b1cfa6365e193ddf5f0#diff-3baf47c64847a8fb8aaa8cc2e088513bL21

dseevr:
Oh, I see, I forgot to populate defaultPaths there.  Got it!

config/archive_test.go
erikh: Hmm. On second thought, I'm not sure this is the right approach. Can we use `contiv/executor` to capture the output?

erikh: we should really avoid setting variables in defer statements, it's really sketch.

config/archive.go
dseevr: I guess this function should not really be directly outputting to stdout but rather returning just returning the info about *what* was changed and letting the creator of the watch decide what to do with it.

config/archive_test.go
dseevr: Sure, I don't think verifying the timestamp is exact really matters for this test.  The TestWatchForPolicyChanges() test below tests that since it has to verify that the stdout output was correct.

config/archive.go
dseevr: Hmm, I guess it technically doesn't matter if a second policy write in the same second doesn't actually generate a revision entry.  Good catch.

config/archive.go
dseevr: I am *never* going to remember `path.Split` exists lol

volcli/volcli.go
dseevr: Originally, no, but based on my other comment, I will change it so that the watcher just emits the data and lets the establisher of the watch decide what to do with it (in this case, print it to stdout).

config/archive.go
erikh: Yeah, this should not be in config, but volcli instead.

config/archive.go
erikh: prevexist is guaranteed atomic.

config/archive_test.go
dseevr: Actually, once the function being tested is modified so it's not directly writing to stdout, this entire test has to change anyways.  We'll just be testing that the watcher emitted what we expected through the channel.

yuva29:
Is there any different ways to change the policy configuration once uploaded rather uploading a brand new policy with the same name?

erikh:
No, the code for that would be extremely complicated and ultimately have only a little benefit. You can specify overrides at `docker create` time via the `--opt` option and the merge tags on each of the structs, e.g.:

```
$ docker volume create --name foo/bar -d volplugin --opt mount=localhost:/mnt
```

yuva29:
Cool. But it doesn't change the underlying policy. sounds fine.

How does policy revision helps to cross-reference with the volume that's created on top of it as stated in issue #269 ? Right now, there is no link between the policy revision and volume. I guess we must be adding some features on top of this change to achieve it.

Please correct me If I'm wrong.

erikh:
They're two discrete features, we should probably put in a second ticket for the other.

yuva29:
Sounds good.

config/archive_test.go
yuva29: Wouldn't it be enough to check that there is some output data from ```ListPolicyRevisions``` ? or
Probably iterate over the result and get policy details using ```GetPolicyRevision``` which will ensure that there is real policy.

dseevr: If we don't check the timestamp in either of the tests, then testing the `TestCreatePolicyRevision()` and `TestListPolicyRevisions()` functions will have identical tests (basically just checking that the # of items being returned has changed by the expected amount).

@yuva29  I like the the idea to use `GetPolicyRevision()`.

I think `TestCreatePolicyRevision()` should test that the created policy entry is under the expected key name (the UNIX timestamp at creation time), and the `ListPolicyRevisions()` should test that all the keys in the response refer to real revisions.  Only one test (i.e., creation) really needs to test that the etcd entry has the correct name.

What do you guys think?

yuva29: I agree with your first point.

Wouldn't this ```c.Assert(revisionCount(s), Equals, 1)``` collide with another policy upload (same policy) by a different host? 

Probably i'm thinking too much here. lol.

yuva29:
What will happen to the revisions if 2 hosts upload the same policy (same timestamp, I guess its pretty much possible) ? Will it override one of them?

config/archive_test.go
dseevr: If there's multiple policy uploads in the same second (whether one or many hosts), the last write wins, but there will only ever be one revision entry for it.

If you added a loop here that created multiple policies and you checked that N revisions were created, the test would fail since they'll all be created in the same second (or some in one second, some in the subsequent second, if the UNIX timestamp crossed a second boundary during the writes).

dseevr:
@yuva29 answered here:  https://github.com/contiv/volplugin/pull/367/files/4200a64bd5181b9d617c2b1cfa6365e193ddf5f0#r72544871

config/archive_test.go
dseevr: Actually, the code for `TestListPolicyRevisions()` and `TestGetPolicyRevision()` would be 95% the same (the only difference is that the latter ensures the policy text matches what was uploaded), so I'm just gonna fold both tests into one.

config/archive_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	resp, err := s.tlc.GetPolicyRevision(testPolicyName, revisions[0])
+	c.Assert(err, IsNil)
+	c.Assert(resp, Equals, policyText)
+}
+
+// this test ensures that a new policy being uploaded results in the correct
+// data (policy name + timestamp, tab-separated) being written to stdout
+func (s *configSuite) TestWatchForPolicyChanges(c *C) {
+	changes := [][2]string{}
```
erikh: missed this before. can we just use `[]string`? Style preference and easier to modify later.

config/archive.go
```diff
@@ -0,0 +1,73 @@
...
+		return "", errors.EtcdToErrored(err)
+	}
+
+	return resp.Node.Value, nil
+}
+
+// WatchForPolicyChanges creates a watch which prints the policy name and revision number
+// (timestamp) to stdout whenever a policy is uploaded.
+func (c *Client) WatchForPolicyChanges(ch chan *watch.Watch, f func(resp *client.Response, w *watch.Watcher)) {
+	w := watch.NewWatcher(ch, c.prefixed(rootPolicyArchive), f)
```
erikh: it would be nice if this did more than provide a facade for the watch lib. Perhaps it should unmarshal the policies, or otherwise fill the `Config` variable.

sorry I missed this on the first pass. I just think watch is being misused here. This is more of what I had in mind for this library: https://github.com/contiv/volplugin/blob/master/config/volume.go#L292-L318

volcli/volcli.go
```diff
@@ -286,6 +289,123 @@ func policyList(ctx *cli.Context) (bool, error) {
...
+
+	cfg.WatchForPolicyChanges(wc, func(resp *client.Response, w *watch.Watcher) {
+		w.Channel <- &watch.Watch{Key: resp.Node.Key, Config: true}
+	})
+
+	for w := range wc {
+		keyspace, revision := path.Split(w.Key)
+		keyspace = keyspace[:len(keyspace)-1] // remove trailing slash
+
+		_, name := path.Split(keyspace)
```
erikh: ok, this one in particular confuses me.

If the path is `/volplugin/policy-archives/policy1/test/<timestamp>`, this will yield the name as `test` not `policy1/test` if I'm reading this right. Is that accurate? If so, we need to change it to `policy1/test`.

If that's how this works disregard, I just don't have bandwidth to test the patch myself.

dseevr: Correct me if I am wrong, but policies are always just a bare name, like `foo`.  Something of the format `foo/bar` would be policy = foo, volume name = bar, right?

`volcli policy upload policy1 < /testdata/ceph/policy1.json` results in `/volplugin/policy-archives/foo/$timestmap` being created.

If we want to support policy names with a slash in them, I can just chop off the prefix, use `path.Split()`, and assume the first value returned is the policy name.  That might be a bit more robust/futureproof.

dseevr: Looks like we don't actually support uploading policies with a slash in them (volcli gets a 404 back), but I flipped the logic here to just chop off the prefix and use `path.Split()`:  https://github.com/contiv/volplugin/pull/367/commits/0a01001e2fd093193632fd414bd07d792c0fa4c0#diff-fa43fdf657b1ed000108d8be7501ee81R70

config/archive.go
dseevr: You didn't miss it.  It wasn't there on your first pass.  :)

It's changed now:  https://github.com/contiv/volplugin/pull/367/commits/0a01001e2fd093193632fd414bd07d792c0fa4c0#diff-fa43fdf657b1ed000108d8be7501ee81R70

config/archive_test.go
dseevr: Done.  I was thinking in terms of a "changeset" being returned as a tuple, but, as long as we're asserting the length is correct, it doesn't really matter.

config/archive_test.go
dseevr: "Does not exist" errors are raised here in a few cases, e.g., when verifying that a policy has zero revisions, if the policy hasn't been uploaded yet there is no directory, so etcdclient returns an `errored.Error`.  

I see some assertions which check that the error text matches some expected value.  I could do that, or I could just defer/recover and return 0.

What is the idiomatic way to handle this?

volcli/volcli.go
erikh: yeah, sorry. this is what I get for replying at 4am. Very sorry.

yuva29:
LGTM

erikh:
LGTM

PR #369, opened 2016-07-27
Fix multi pool snapshots

### LINKED ISSUE #362: [volsupervisor] rbd clone invocation incorrect for non-default pools
ceph version: 0.94.6
volplugin version: 0.20.0

with a pool named "example", the following invocation of volcli:

```
$ volcli  volume snapshot copy policy1/volume1 2016-07-22-22:30:00.76904695-+0000-UTC  volume1-clone1
```

invokes:

```
$ rbd clone policy1.volume1 policy1.volume1-clone1 --snap 2016-07-22-22:30:00.76904695-+0000-UTC --pool example
rbd: error opening pool rbd: (2) No such file or directory
```

this results in a  500 error from the API server and the clone isn't created.  the correct rbd invocation is:

```
rbd clone example/policy1.volume1 example/policy1.volume1-clone1 --snap 2016-07-22-22:30:00.76904695-+0000-UTC
```

the first `example/` is the source pool and the second is the destination pool. the `--pool` flag isn't necessary.

erikh:
Ceph's pool/image syntax is now used. This corrects many errors taking snapshots from alternative pools, which previously attempt to use `--pool` which unfortunately is not supported for some snapshot operations. The pool syntax has been unified in the call to mkpool().

Fixes #362

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -2,6 +2,7 @@ package ceph
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -39,6 +40,21 @@ var spaceSplitRegex = regexp.MustCompile(`\s+`)
 //
 type Driver struct {
 	mountpath string
+}
+
+func (c *Driver) mkMountPath(poolName, intName string) (string, error) {
+	// Directory to mount the volume
+	volumePath := filepath.Join(c.mountpath, mkpool(poolName, intName))
+	rel, err := filepath.Rel(c.mountpath, volumePath)
+	if err != nil || strings.Contains(rel, "..") {
+		return "", errors.MountFailed.Combine(errored.Errorf("Calculated volume path would escape subdir jail: %v", volumePath))
+	}
+
+	return volumePath, nil
+}
+
+func mkpool(poolName, volumeName string) string {
+	return fmt.Sprintf("%s/%s", poolName, volumeName)
 }
 
 func runWithTimeout(cmd *exec.Cmd, timeout time.Duration) (*executor.ExecResult, error) {
@@ -99,7 +115,7 @@ func (c *Driver) Create(do storage.DriverOptions) error {
 		return err
 	}
 
-	cmd := exec.Command("rbd", "create", intName, "--size", strconv.FormatUint(do.Volume.Size, 10), "--pool", do.Volume.Params["pool"])
+	cmd := exec.Command("rbd", "create", mkpool(do.Volume.Params["pool"], intName), "--size", strconv.FormatUint(do.Volume.Size, 10))
 	er, err := runWithTimeout(cmd, do.Timeout)
 
 	if er != nil {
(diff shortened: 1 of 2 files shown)


erikh:
@yuva29 @dseevr PTAL

storage/backend/ceph/ceph.go
```diff
@@ -41,6 +42,21 @@ type Driver struct {
 	mountpath string
 }
 
+func (c *Driver) mkMountPath(poolName, intName string) (string, error) {
+	// Directory to mount the volume
+	volumePath := filepath.Join(c.mountpath, mkpool(poolName, intName))
+	rel, err := filepath.Rel(c.mountpath, volumePath)
+	if err != nil || strings.Contains(rel, "..") {
+		return "", errors.MountFailed.Combine(errored.Errorf("Calculated volume path would escape subdir jail: %v", volumePath))
```
dseevr: Security-wise, will we ever have to worry about symlink resolution in mount targets?

dseevr:
LGTM

storage/backend/ceph/ceph.go
erikh: I'll have to test but mounting through a symlink probably causes EINVAL

erikh: I'll get back to you on this.

dseevr: At least for bind mounts, symlinks are resolved during mounts:

```sh
root:/tmp$ mkdir realdir
root:/tmp$ touch realdir/data.txt
root:/tmp$ ln -s realdir symlink
root:/tmp$ mkdir mountpoint
root:/tmp$ mount --bind /tmp/symlink /tmp/mountpoint/
root:/tmp$ ls -l /tmp/mountpoint/
data.txt
```

I have no idea if it applies at all to what we're doing, but I was curious.

erikh: I'll look into this, we need the ceph device to exhibit this behavior.

I'm also not 100% certain it matters because we'll be overwriting the dir with the mounted content anyway.

dseevr: Cool, thanks.

storage/backend/ceph/ceph_test.go
```diff
@@ -109,6 +128,20 @@ func (s *cephSuite) TestMountUnmountVolume(c *C) {
...
+		Timeout:   5 * time.Second,
+	}
+
+	if done {
+		return
+	}
+
+	done = true
+
+	goto again
```
yuva29: In case of successful completion during the first run, why are we preforming another loop of execution rather returning? 

What is the actual intention of using goto here?

yuva29: Does it actually perform 2 rounds of execution with different volume specification(volumeSpec, volumeSpecTestPool ? And is it why the function doesn't return on first successful completion?

erikh: yes, it performs 2 rounds. there is no need for return in the test functions -- that doesn't accomplish anything. `c.Assert` will stop the routine if anything fails.

yuva29:
LGTM

erikh:
Leaving this open until I get the opportunity  to test the mount/symlink semantics.

erikh:
Ok I looked into this. I found an exploit but it's hard to trigger. I'll be putting in an issue for it anyways.

In summary: 

Since the data will always originate from the RBD or NFS mount or etc, this will ensure that no data from the host can leak into the container. That's good.

The bad: if I were to manage this:

`ln -s / /path/to/mountpoint`

I could effectively mount over the whole root filesystem.

NOW

both of these strategies rely on being able to put a symlink on the host. I'm not very worried about this as if the host is compromised, well...

So I'm not going to make this very high priority for now. If there is an easier to exploit or pure remote exploit we can offer here, I would definitely raise the priority.

PR #372, opened 2016-07-28
Consul

erikh:
This is nearly a copy of https://github.com/contiv/ansible/pull/255. It also includes an upgrade of our base ansible set as well.


dseevr:
LGTM

PR #373, opened 2016-07-29
Godeps: bump remotessh to the latest known commit

unclejack:
This PR vendors a known version of remotessh. This is the current version of this dependency. I'll bring in some changes to avoid this in the future.


erikh:
this is fixed in #366 which is scheduled to be merged today; since it needs additional work and requires a rebase if I merge this I'm going to close this ticket in favor of that one.

Issue #358, opened 2016-07-19
tests with heisenbugs


erikh:
Going to keep a running list of these as I encounter them in CI. Please take one or two if you get time.

* [ ] `systemtestSuite.TestVolsupervisorRestart`
* [ ] `systemtestSuite.TestBatteryParallelMount` (nfs only)

yuva29:
- [ ] `systemtestSuite.TestBatteryMultiMountSameHost `

erikh:
I think I fixed the multi mount problems in #374, can you review again when merged?

erikh:
these tests have been resolved.

PR #374, opened 2016-07-30
Docker API abstraction

erikh:
This set of interfaces (api/interfaces.go) allows us to project volplugin as a
conduit for any scheduler to schedule mounts. While not here yet, this is a
prerequisite to kubernetes, nomad, and native mesos support.

Many changes have been made to accomodate this. Too many to split into separate
commits at this point. Instead, I will itemize some of them below:

* Mount management in volplugin has drastically changed (for the better). A
  number of panics and other anomalies related to races at boot time were fixed
  as a result of this change. These are in api/internals/mount and volplugin/runtime.go.
* Mount reporting in NFS vastly improved. This logic is located in
  storage/backend/nfs/mount.go.
* As a result of the rewrite, a new umbrella type for volumes (api.Volume) has
  been introduced to attempt to unify all the other volume types in the system.
  It should be used for anything that will write to DB or the API service(s).
  storage.Volume still exists for storage backends, however.
* A number of timing issues (heisenbugs) with systemtests were resolved due to
  the rewrite exposing the timing issues.


erikh:
I'm checking this in now; it's ready for review but may still need a few fixups to pass CI.

erikh:
One last thing; this is lacking a lot of tests atm. I plan on filling those in late next week after the consul stuff is done.

api/api.go
```diff
@@ -1,187 +1,136 @@
...
 
-	if policy == "" {
-		a.HTTPError(w, errors.GetPolicy.Combine(errored.Errorf("policy was blank for volume %q", req.Name)))
-		return
-	}
+// Action is a catchall for additional driver functions.
+func Action(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+	log.Debugf("Unknown driver action at %q", r.URL.Path)
+	content, _ := ioutil.ReadAll(r.Body)
```
dseevr: A successful call here returns `err == nil`, so we should check the error.

erikh: yeah... this is very old code. :)

api/api.go
```diff
@@ -1,187 +1,136 @@
...
-		return
-	}
-
-	log.Errorf("Returning HTTP error handling plugin negotiation: %s", err.Error())
-	http.Error(w, string(content), http.StatusOK)
+// AddStopChan adds a stop channel for the purposes of controlling mount ttl refresh goroutines
+func (a *API) AddStopChan(name string, stopChan chan struct{}) {
+	a.lockStopChanMutex.Lock()
+	a.lockStopChans[name] = stopChan
+	a.lockStopChanMutex.Unlock()
```
dseevr: This Unlock() should be defer'ed to ensure that it gets called even in a panic() situation

api/api.go
```diff
@@ -1,187 +1,136 @@
...
-
-// Action is a catchall for additional driver functions.
-func Action(w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-	log.Debugf("Unknown driver action at %q", r.URL.Path)
-	content, _ := ioutil.ReadAll(r.Body)
-	log.Debug("Body content:", string(content))
-	w.WriteHeader(503)
+	delete(a.lockStopChans, name)
+	a.lockStopChanMutex.Unlock()
```
dseevr: Same with this one

api/api.go
erikh: the daemon becomes unresponsive after a panic, I don't think this is necessary.

erikh: the whole point of the panic is to terminate the program immediately to avoid causing damage.

dseevr: Cool, thanks for the context.

api/api.go
```diff
@@ -1,187 +1,136 @@
...
+	driverOpts := storage.DriverOptions{}
+	volConfig, err := a.Client.GetVolume(uc.Policy, uc.Name)
 	if err != nil {
-		a.HTTPError(w, errors.GetHostname.Combine(err))
-		return
+		return nil, nil, driverOpts, err
 	}
 
-	policyObj, err := a.Client.GetPolicy(policy)
+	driver, err := backend.NewMountDriver(volConfig.Backends.Mount, (*a.Global).MountPath)
```
yuva29: You might want to use `NewDriver()` here. In that way, we will not expose `NewMountDriver`, `NewSanpshotDriver`, etc.

erikh: I'll look into this; that patch wasn't here before.

api/api.go
yuva29: In that case after panic if an other container in the same host wants to get this lock, it will be blocked right because the panicked  process didn't unlock it ?

dseevr: I think, as a general pattern, we should try to defer mutex unlocking when possible because this makes it safe for other people to call this function again if they're recovering from panics.  I think this is especially true for exported functions.

If we do defer an unlock in our function, it doesn't affect us at all if we're not recover()ing.  If we don't defer and the caller recover()s from a panic in our function, the mutex is still locked.  A small change on our part makes a function safe for all future users of it.

Examples:  https://play.golang.org/p/yDp7JOvqCy

In this specific function we're commenting on, I don't really think a map assignment is going to panic... but I'm just speaking in general.  Using defer is more idiomatic Go, too.

api/handlers.go
```diff
@@ -0,0 +1,333 @@
...
+	a.HTTPError(w, err)
+	return
+}
+
+func (a *API) getMountPath(driver storage.MountDriver, driverOpts storage.DriverOptions) (string, error) {
+	path, err := driver.MountPath(driverOpts)
+	return path, err
+}
+
+// Path is the handler for both Path and Remove requests. We do not honor
```
dseevr: Can you explain what is meant by "we do not honor remove requests" here?  What happens if one is sent in?

api/api.go
yuva29: ah. Unless otherwise we recover(after unlocking), none of the containers can execute anything. Makes sense!

Thanks for clarifying @dseevr

yuva29: Just one more thing here, how do we recover from panic if there is one ? I couldn't find any recovery code in volplugin.

api/handlers.go
erikh: nothing happens. volcli is used to remove volumes.

api/api.go
erikh: they are and will not be recovering from panics. no code in volplugin will do that.

the reason is the daemon *SHOULD NEVER* panic unless there is a critical, unresolvable error. The process should always auto-heal if it can. Panicking is a last resort.

dseevr: @yuva29 Since Go lacks normal exception trapping and doesn't have RAII + destructors or anything like that, the only way to guarantee that a resource is freed after being allocated is to immediately defer the cleanup/destroy/free function.

Sometimes you may not want to do this, e.g., in a certain function, you want to lock, do a quick operation, unlock, then do a long running operation not dependent on the lock.  In that case, you should move the "lock, do operation, unlock" code into a separate function and call it that way you can guarantee the cleanup happens even if the "do operation" step panics.  You could also just do it inline with an anonymous func:

```go
func doStuff() {
    func(){
        foo.Lock()
        defer foo.Unlock()
        someProtectedCall()
    }

    someLongRunningCall()
}
```

There's really no other way around it, unfortunately.

erikh: oh this is a different routine. Ok, this Unlock defer is not required because the above code will never panic. I'm not sure if that's obvious or not.

It's literally just more stack overhead, there's no point.

erikh: another way of putting this: in the extremely unlikely event the panic does happen and it doesn't make it to main (which will abort the program), I'd rather have that mutex hang everything instead of it taking further action. Something is very f'd up if it does.

dseevr: @erikh Got it, thanks.

yuva29: okay, that helps. Thanks for the clarification @erikh @dseevr

storage/backend/nfs/mount.go
```diff
@@ -0,0 +1,72 @@
...
+			continue
+		}
+
+		parts := strings.Split(line, " ")
+		devParts := strings.Split(parts[2], ":")
+		if len(devParts) != 2 {
+			return nil, errored.Errorf("Could not parse %q properly.", mountInfoFile)
+		}
+
+		if devParts[0] == nfsMajorMountType && parts[8] == "nfs4" {
```
dseevr: How do you feel about storing these parts[]/devParts[] into named variables?  It'd make this code a *lot* more readable IMO.  For example, `majorId` communicates a lot more than `devParts[0]`, and you don't have to repeat the array indexes multiple times, which is getting close to magic number territory.

Having the variables all grouped together would allow someone, at a glance, to see what the important parts of the chopped up string actually are.

dseevr: Maybe something like this:  https://github.com/dseevr/volplugin/blob/docker-api-mount-changes/storage/backend/nfs/mount.go#L41

(I think I got the variable names right lol)

systemtests/volplugin_test.go
```diff
@@ -92,30 +96,33 @@ func (s *systemtestSuite) TestVolpluginCrashRestart(c *C) {
...
 	c.Assert(stopVolplugin(s.vagrant.GetNode("mon0")), IsNil)
+	time.Sleep(time.Second)
 	c.Assert(startVolplugin(s.vagrant.GetNode("mon0")), IsNil)
 	c.Assert(waitForVolplugin(s.vagrant.GetNode("mon0")), IsNil)
+	time.Sleep(30 * time.Second)
 
-	out, err := s.volcli(fmt.Sprintf("volume runtime upload %s < /testdata/iops1.json", volName))
+	out, err = s.volcli(fmt.Sprintf("volume runtime upload %s < /testdata/iops1.json", volName))
 	c.Assert(err, IsNil, Commentf(out))
-	time.Sleep(45 * time.Second)
```
dseevr: Why was this sleep able to be decreased by such a large amount?  Very cool.  😃

dseevr: Oh, nvm... I see 35 seconds was added back in two Sleep() calls above.

volplugin/init.go
```diff
@@ -2,67 +2,107 @@ package volplugin
...
+	// This is intentional to ensure we don't take any action when docker is down.
+	for {
+		dockerClient, err := client.NewEnvClient()
+		if err != nil {
+			return nil, nil, errored.Errorf("Could not initiate docker client").Combine(err)
+		}
+
+		containers, err := dockerClient.ContainerList(context.Background(), types.ContainerListOptions{})
+		if err != nil {
+			if now.Sub(time.Now()) > dc.Global.Timeout {
```
dseevr: Can we use `time.Since(now)` here?

volplugin/init.go
```diff
@@ -2,67 +2,107 @@ package volplugin
...
 
-		for _, mount := range mounts {
-			parts := strings.Split(mount.Volume.Name, "/")
+	return mounts, counts, nil
+}
+
+func (dc *DaemonConfig) updateMounts() error {
+	// this loop is used to ensure that volplugin itself has a clear understanding of what's going on.
+	//
+	// FIXME If volplugin detected no mount but docker thinks there is a volplugin
```
dseevr: Just for my (and maybe @yuva29's) edification, what steps can result in volplugin/docker being in this situation?

dseevr:
Aside from my pedantry, LGTM

config/config.go
```diff
@@ -27,14 +28,18 @@ const (
 
 var defaultPaths = []string{rootVolume, rootUse, rootPolicy, rootPolicyArchive, rootSnapshots}
 
-// Request provides a request structure for communicating with the
-// apiserver.
-type Request struct {
-	Volume  string            `json:"volume"`
+// VolumeRequest provides a request structure for communicating volumes to the
+// apiserver or internally. it is the basic representation of a volume.
+type VolumeRequest struct {
```
yuva29: I guess we have to move this struct to docker/structs.go

volplugin/init.go
erikh: sure

volplugin/init.go
erikh: volplugin loses state of the mount (can't find it on boot) but docker still thinks it's mounted. This can happen if volplugin's internal state corrupts, or if the volume is unmounted on the host (it will still be mounted in the container, but not available to the host). I don't want to restore from external state (e.g., a DB) because that just adds another thing that can go wrong.

api/api.go
erikh: I don't want to sound rude, but how hard is it to `cat /proc/self/mountinfo` and figure this out?

erikh: I mean, you clearly did it. :D

erikh: fucking github, wrong thread.

storage/backend/nfs/mount.go
erikh: I don't know why this comment didn't show up here last time, but...
`cat /proc/self/mountinfo` (which I think is very clearly spelled out at the top of the function) is the file format. For editing's sake this is simpler and I don't see the value personally when you can just cat the file to display this information.

config/config.go
erikh: This is related to the config library (where database operations are performed) the structs in the docker library are just API exchange requests.

yuva29: Oh, sorry. I was confused by the name.

api/api.go
dseevr: Yes, I did do that 😄  but I wouldn't have had to + repeated bare numerical constants is kind of a code smell IMO

yuva29:
Can you please explain how is this prevented here https://github.com/contiv/volplugin/issues/358

erikh:
Sure, the locking code has changed a bit. You can see that some of the locking code was reorganized (so that certain locks are taken sooner) and the boot-time population of state is different.

api/api.go
erikh: Ok since this code is partially duplicated, I've created this issue to resolve it: https://github.com/contiv/volplugin/issues/381

For now I'm going to leave this alone because #381 should really be the way to fix this... and this code works today.

erikh:
Pushed latest revision. Merging after tests pass.

To address open issues:

* @yuva29 and I have determined the NewDriver() method needs some work. #382 was created to address this.
* The mountscan issue is going to be resolved by #381. @yuva29 is currently working on this.

PR #375, opened 2016-07-31
Speedup build

erikh:
This fixes a golint bug + introduces a better / faster / more robust way of checking our code via the `build/scripts/checks.sh` script, which you can run inside the vms with `make checks`.


unclejack:
LGTM

dseevr:
I like the Makefile cleanup 👍

PR #376, opened 2016-08-01
Vagrantfile: disable GRO on VirtualBox

unclejack:
This PR works around some of the VMM bugs discovered in VirtualBox 5.0.x and fixed in 5.1.x.


erikh:
We're going to be on 5.1 in a week or two at the latest, is this a necessary change?

unclejack:
@erikh I expect this to be an improvement even with 5.1.

erikh:
According to jenkins build times this took 15m less than the average. Merging.

I also have a fix that makes vbox 5.1 work.

PR #380, opened 2016-08-02
Unit-test for JSON validation

yuva29:
Unit test for all the functions in config/validation.go

### DIFF
--- a/config/validation_test.go
+++ b/config/validation_test.go
@@ -1,0 +1,329 @@
+package config
+
+import . "gopkg.in/check.v1"
+
+var (
+	defaultBackends = &BackendDrivers{CRUD: "ceph", Mount: "ceph", Snapshot: "ceph"}
+	VolumeConfigs   = map[string]map[string]*Volume{
+		"valid": {
+			"basic": {
+				DriverOptions:  map[string]string{"pool": "rbd"},
+				CreateOptions:  CreateOptions{Size: "10MB"},
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+				VolumeName:     "basicvolume",
+				PolicyName:     "basicpolicy",
+				Backends:       defaultBackends,
+			},
+			"basicwithruntime": {
+				DriverOptions:  map[string]string{"pool": "rbd"},
+				CreateOptions:  CreateOptions{Size: "10MB"},
+				RuntimeOptions: RuntimeOptions{UseSnapshots: true, Snapshot: SnapshotConfig{Frequency: "10m", Keep: 10}},
+				VolumeName:     "basicvolume",
+				PolicyName:     "basicpolicy",
+				Backends:       defaultBackends,
+			},
+		},
+		"invalid": {
+			"emptyvolume": {}, // base case
+			"emptyname": {
+				VolumeName:     "",
+				PolicyName:     "policy",
+				Backends:       defaultBackends,
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+			"emptypolicy": {
+				VolumeName:     "volume",
+				PolicyName:     "",
+				Backends:       defaultBackends,
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+			"emptybackends": {
+				VolumeName:     "volume",
+				PolicyName:     "policy",
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+			"invalidnames": {
+				VolumeName:     "volume/1.2",
+				PolicyName:     "policy/1/2",
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+			"invalidvolumename": {
+				VolumeName:     "volume.1",
+				PolicyName:     "policy",
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+			"invalidpolicyname": {
+				VolumeName:     "volume",
+				PolicyName:     "policy.1/2",
+				RuntimeOptions: RuntimeOptions{UseSnapshots: false},
+			},
+		},
+	}
+	PolicyConfigs = map[string]map[string]*Policy{
+		"valid": {
+			"basic": testPolicies["basic"],
+			"basicceph": {
+				Name: "basicceph",
+				Backends: &BackendDrivers{
+					CRUD:     "ceph",
+					Mount:    "ceph",
+					Snapshot: "ceph",
+				},
+			},
+			"basicnfs": {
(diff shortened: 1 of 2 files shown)


config/validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+				UseSnapshots: true,
+				Snapshot: SnapshotConfig{
+					Keep:      1,
+					Frequency: "", // "" not allowed
+				},
+			},
+			"invalidfrequency": {
+				UseSnapshots: true,
+				Snapshot: SnapshotConfig{
+					Frequency: "badvalue", // accpets only (^[0-9]+.$)
```
dseevr: Type: accpets -> accepts

dseevr: One other thing to be aware of with regex, ^ and $ are not "start of input" and "end of input".  You'll want to use `\A` and `\z` (lower case z... `\Z` is different!) for start/end of input.

If you use something like `/^[0-9]+$/`, a string like `"12345\nwheee"` will pass the validation!

`\Z` allows an optional trailing newline.  `/\A[0-9]+\Z/` will match `"12345"` and `"12345\n"`!

Sorry if I missed this in the other PR...  I'm usually on top of this especially in web apps because it's an attack vector and can have disastrous results.

config/validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func (s *configSuite) TestSingletonBackend(c *C) {
+	PolicyConfigs["valid"]["backendceph"].Validate() // Policy Validation
+	s.validateBackendsConfig(c, PolicyConfigs["valid"]["backendceph"].Backends, "ceph", "ceph", "ceph")
+
+	PolicyConfigs["valid"]["backendnfs"].Validate()
+	s.validateBackendsConfig(c, PolicyConfigs["valid"]["backendnfs"].Backends, "", "nfs", "")
+
+	// Below test ensures that "Validate" didnot change the given "backends" config, incase if there is one provided
```
dseevr: typo: didnot -> did not

config/validation_test.go
dseevr: Yes, I did miss this... sorry.

https://github.com/yuva29/volplugin/blob/b7f230ff31dacbcae28007453d4db340c21c5326/config/schema.go#L14

erikh: regexp engines are different between different systems. oniguruma is ruby's engine, which is very similar to perl's. Both are non-deterministic.

RE2, however.... which is the golang implementation, is different. It's a deterministic regexp engine. In this case it's the same, but in several cases you'd be incorrect (like egrep, or sed, or emacs).

Anyways my point. Specifically from https://golang.org/pkg/regexp/syntax/:

```
^              at beginning of text or line (flag m=true)
$              at end of text (like \z not \Z) or line (flag m=true)
\A             at beginning of text
\b             at ASCII word boundary (\w on one side and \W, \A, or \z on the other)
\B             not at ASCII word boundary
\z             at end of text
```

`\A` and `\z` only differ from `^` and `$` when `m` is applied. I think you'll find that this behavior is exactly the same in perl and ruby, but ruby turns `m` on by default.

The goal of this lecturing is to hopefully educate a little bit about how every single regexp implementation is different, and also express that this code should have no issues unless something in between sets `m`.

dseevr: Did not know that about Go's regex engine. Thanks for the info!

Out of curiosity, I just tested `\d` in Go and it surprisingly only matches `[0-9]`.  In some other implementations, it matches like 200+ characters or something.  I'll definitely be making fewer assumptions about Go regexes in the future. 😄

dseevr:
LGTM

I think you need to amend and force push to trigger another CI build, though.  The current build died with an "etcd is not healthy" error.

config/validation_test.go
yuva29: Erik, nothing sets `m`. go check just adds `^` & `$` 
https://github.com/yuva29/volplugin/blob/master/vendor/gopkg.in/check.v1/checkers.go#L307

So this will pass.

dseevr: Yeah, looks like your original implementation was fine.  Sorry for the pointless tangent. 😕

erikh: Yeah go's regexp engine is very bare-bones. Be sure to check the docs when using it.

config/validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+				VolumeName:     "basicvolume",
+				PolicyName:     "basicpolicy",
+				Backends:       defaultBackends,
+			},
+		},
+		"invalid": {
+			"emptyvolume": {}, // base case
+			"emptyname": {
+				VolumeName:     "",
+				PolicyName:     "policy",
```
erikh: add a test case in this area for names with a / in them (policy or volume). that is also invalid.

erikh: also check for `.`

sorry for these late-comers but I just realized that we can do this right here now instead of all over the place.

yuva29: What about other special characters? Is it allowed?

config/validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (s *configSuite) validateBackendsConfig(c *C, backends *BackendDrivers, crud string, mount string, snapshot string) {
+	c.Assert(backends.CRUD, Equals, crud)
+	c.Assert(backends.Mount, Equals, mount)
+	c.Assert(backends.Snapshot, Equals, snapshot)
+}
+
+func (s *configSuite) TestPolicyValidation(c *C) {
+	for _, policy := range PolicyConfigs["valid"] {
+		c.Assert(policy.ValidateJSON(), IsNil)
```
erikh: Let's do the same here as that comment above.

config/validation_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	err = invalidPolicyConfigs["nfsbackends"].ValidateJSON()
+	c.Assert(err, ErrorMatches, "(?m)*backends.crud must be one.*")
+	c.Assert(err, ErrorMatches, "(?m)*backends.snapshot must be one.*")
+}
+
+//XXX: All the possible runtime configs are tested in TestRuntimeValidation, so I'm not repeating it here.
+func (s *configSuite) TestVolumeValidation(c *C) {
+	for _, volume := range VolumeConfigs["valid"] {
+		c.Assert(volume.ValidateJSON(), IsNil)
```
erikh: here too.

erikh:
This patch looks great. Let's tidy up the last bits and merge it.

erikh:
BTW, major props on the error checking; very thorough.

erikh:
LGTM

PR #383, opened 2016-08-02
api: fix the log message for creating a volume: the new String() was …

erikh:
…causing it to accept an invalid argument.


yuva29:
LGTM

dseevr:
LGTM

PR #384, opened 2016-08-02
Makefile,build/scripts: validate deps from vendor

unclejack:
Godep makes it possible to vendor a locally modified dependency. This isn't a major issue when it comes to the properly functioning code which depends on it. This is a major issues when trying to restore the deps using `godep restore` and when trying to bump a dependency.

This PR adds a script which makes it possible to check that `godep restore` doesn't rely on any locally modified deps. This lets us avoid the issue of having a local commit hash added to Godeps.json and the files of that particular commit added to vendor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,9 @@ ssh:
 checks:
 	vagrant ssh mon0 -c "sudo -i sh -c 'cd $(GUESTGOPATH); ./build/scripts/checks.sh'"
 
+vendor-check:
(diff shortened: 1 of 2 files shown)


erikh:
`godep restore` never takes anything from the local system, it takes it from Godeps.json. I'm not sure what the goal is here.

Additionally I don't like the idea of having yet another place we track ignored files. We have a lot of them already.

unclejack:
@erikh Godep will happily vendor from GOPATH if you've committed something locally on a branch. This made it possible to ensure commits such as https://github.com/contiv/volplugin/pull/366/commits/4f46dd903e4bae62cc50c5a1db7aca7c46aaa416 will not be required again.

PR #386, opened 2016-08-04
api: fix some issues with the lock handler. See below.

erikh:
A number of things changed:

* Use locks are *never* taken in unlock mode. This corrected a number of
  potential (but unproven) races when switching between the two modes.
  Locked/unlocked can theoretically be runtime now.
* Order of operations:
  * TTL refresh is garbage collected
  * the long-term use lock is taken
  * the mount counter is checked
  * mount occurs
  * then the TTL refresh is restarted
* This corrects situations where the TTL refresh would trump the permanent lock

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -203,24 +203,21 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	volName := volConfig.String()
+	if !volConfig.Unlocked {
+		ut := &config.UseMount{
+			Volume:   volName,
+			Reason:   lock.ReasonMount,
+			Hostname: a.Hostname,
+		}
 
-	ut := &config.UseMount{
-		Volume:   volName,
-		Reason:   lock.ReasonMount,
-		Hostname: a.Hostname,
-	}
-
-	if volConfig.Unlocked {
-		ut.Hostname = lock.Unlocked
-	}
-
-	// XXX the only times a use lock cannot be acquired when there are no
-	// previous mounts, is when in locked mode and a mount is held on another
-	// host. So we take an indefinite lock HERE while we calculate whether or not
-	// we already have one.
-	if err := a.Client.PublishUse(ut); err != nil {
-		a.HTTPError(w, errors.LockFailed.Combine(err))
-		return
+		// XXX the only times a use lock cannot be acquired when there are no
+		// previous mounts, is when in locked mode and a mount is held on another
+		// host. So we take an indefinite lock HERE while we calculate whether or not
+		// we already have one.
+		if err := a.Client.PublishUse(ut); err != nil {
+			a.HTTPError(w, errors.LockFailed.Combine(err))
+			return
+		}
 	}
 
 	// XXX docker issues unmount request after every mount failure so, this evens out
@@ -242,13 +239,13 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	stopChan, err := a.Lock.AcquireWithTTLRefresh(ut, (*a.Global).TTL, (*a.Global).Timeout)
-	if err != nil {
-		a.HTTPError(w, errors.LockFailed.Combine(err))
-		return
+	// Only perform the TTL refresh if the driver is in unlocked mode.
+	if !volConfig.Unlocked {
+		if err := a.startTTLRefresh(volName); err != nil {
+			a.HTTPError(w, err)
+			return
+		}
 	}
-
-	a.AddStopChan(volName, stopChan)
 
 	mc, err := driver.Mount(driverOpts)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dseevr:
LGTM

api/handlers.go
```diff
@@ -289,13 +303,42 @@ func (a *API) Unmount(w http.ResponseWriter, r *http.Request) {
...
+		// before taking it so it is not cleared in the unlikely event the mount
+		// takes longer than the TTL. Re-establish the TTL on error only if it is in
+		// locked mode.
+		if err := a.Client.PublishUse(ut); err != nil {
+			a.HTTPError(w, errors.LockFailed.Combine(err))
+			return
+		}
+	}
+
 	if a.MountCounter.Sub(volName) > 0 {
```
yuva29: Ideally, this holds true only for `unlocked` volumes. then why do we need L335 loop here? 

Please correct me if I'm wrong.

api/handlers.go
```diff
@@ -289,13 +303,42 @@ func (a *API) Unmount(w http.ResponseWriter, r *http.Request) {
...
+		// anything that doesn't acquire a lock. We also remove the TTL refresh
+		// before taking it so it is not cleared in the unlikely event the mount
+		// takes longer than the TTL. Re-establish the TTL on error only if it is in
+		// locked mode.
+		if err := a.Client.PublishUse(ut); err != nil {
+			a.HTTPError(w, errors.LockFailed.Combine(err))
+			return
+		}
+	}
+
```
yuva29: Ideally, L327 holds true only for `unlocked` volumes. then why do we need L335 loop here? 

Please correct me if I'm wrong.

api/handlers.go
```diff
@@ -289,13 +303,42 @@ func (a *API) Unmount(w http.ResponseWriter, r *http.Request) {
...
+	if !volConfig.Unlocked {
+		a.RemoveStopChan(volName)
+
+		// XXX to doubly ensure we do not UNMOUNT something that is held elsewhere
+		// (presumably because it is mounted THERE instead), we refuse to unmount
+		// anything that doesn't acquire a lock. We also remove the TTL refresh
+		// before taking it so it is not cleared in the unlikely event the mount
+		// takes longer than the TTL. Re-establish the TTL on error only if it is in
+		// locked mode.
+		if err := a.Client.PublishUse(ut); err != nil {
```
yuva29: Incase of locked mount(only one mount exists), this should be the last mount then why do we need to publish the use again?  I really don't get this one. Please explain me.

erikh: I forgot to detail this; nice catch.

Basically, when we act on UNMOUNT we want to be sure the lock still exists, or if not, that it's not taken. This is to ensure we don't have a situation where the lock is trumped during unmount, which I observed a few days back. We already know via docker that the volume should already be mounted.

api/handlers.go
erikh: so there are two kinds of locks in this discussion: permanent and TTL-bound locks.

The first `PublishUse` call takes a permanent lock. It will not expire.

The `startTTLRefresh` will enable a goroutine to periodically refresh the key with a TTL expiry. This ensures that if volplugin or the host crashes, that the mount will eventually expire.

api/handlers.go
```diff
@@ -289,13 +303,42 @@ func (a *API) Unmount(w http.ResponseWriter, r *http.Request) {
...
 	volName := volConfig.String()
 
+	ut := &config.UseMount{
+		Volume:   volName,
+		Reason:   lock.ReasonMount,
+		Hostname: a.Hostname,
+	}
+
+	if !volConfig.Unlocked {
+		a.RemoveStopChan(volName)
```
yuva29: Can you please explain me the use of `a.AddStopChan(volName, stopChan)` and `a.RemoveStopChan(volName)`. I couldn't get much from the code.

erikh: the stop channel (or `stopChan`) is the channel, when sent an empty struct, terminates the TTL refresher.

`AddStopChan` adds a stop channel to a registry which is keyed by volume name. any additional adds will terminate the old refreshers.
`RemoveStopChan` terminates the goroutine and removes the stop channel from the map.

api/handlers.go
```diff
@@ -289,13 +303,42 @@ func (a *API) Unmount(w http.ResponseWriter, r *http.Request) {
...
+
 	if a.MountCounter.Sub(volName) > 0 {
 		log.Warnf("Duplicate unmount of %q detected: ignoring and returning success", volName)
 		path, err := a.getMountPath(driver, driverOpts)
 		if err != nil {
 			a.HTTPError(w, errors.MarshalResponse.Combine(err))
 			return
 		}
+
+		if !volConfig.Unlocked {
```
yuva29: @erikh Is there a way to check if this particular path being covered in some test case? 

I doubt if this block will execute ever.

api/handlers.go
yuva29: okay. Now, I understand that the whole purpose of this loop is to ensure that lock is still there. 

If so, Isn't enough just to check if publishUse fails on "ErrExists"

api/handlers.go
erikh: I can assure you it's executed in locked mode; everything would hang in the tests otherwise.

The great thing about the mount/unmount code is that it's easy to test in systemtests. nothing will work if it's broken.

The bad thing about the mount/unmount code is that it all requires multi-host functionality, and our UNIT tests are not sufficient to test this behavior.

api/handlers.go
erikh: it isn't a loop, to be clear. it never repeats.

the purpose of hte lock is to acquire it so nothing else can.

yuva29:
Code LGTM. 

You can merge this now. We can talk about it further offline.

erikh:
sgtm. merging

PR #387, opened 2016-08-04
config: remove another noisy log message

erikh:

### DIFF
--- a/config/validation.go
+++ b/config/validation.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/contiv/errored"
 
-	log "github.com/Sirupsen/logrus"
-
 	gojson "github.com/xeipuuv/gojsonschema"
 )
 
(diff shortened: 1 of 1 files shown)


yuva29:
Sorry for all those log messages. I thought it will somehow help us in debugging incase of an issue.

erikh:
Don't feel bad about this.

Code review is to catch the things we don't want. Both Bill and I missed
these and it's not a huge deal that we did, but we did.

We fixed it. This will happen very frequently in open source, you'll write
something and someone will come in and touch it up.

Point is: It's a team effort. We failed to correct the error the first
time, so we're correcting it now.

FWIW, if you want to circle back and add Debug logging with a little more
information in the log message that'd be perfect.

erikh:
since this is a log message and ansible failed to provision on one of the tests, I'm just going to merge this.

PR #390, opened 2016-08-05
Unify mountscan across all backends

### LINKED ISSUE #381: Unify mountscan code in storage drivers
This should live under `storage/mountscan` and be a merge of the mountscan code in ceph and nfs storage drivers which essentially do the same thing, just tweaked a bit.

What I would like this new code to do is parse the `/proc/self/mountinfo` text file into a structure which can then be evaluated by any storage driver for any need.

yuva29:
Fixes #381


storage/mountscan/mountscan.go
```diff
@@ -0,0 +1,206 @@
...
+	deviceInfoFile          = "/proc/devices"
+	nfsMajorID              = 0
+	totalMountInfoFieldsNum = 11
+)
+
+// GetMountsRequest captures all the params required for scanning mountinfo
+type GetMountsRequest struct {
+	DriverName   string // ceph, nfs
+	FsType       string // nfs4, ext4
+	KernelDriver string // rbd, device-mapper, etc.
```
erikh: I'm not sure I like this, largely because it's a member only used for one thing.

erikh:
I'll pour over this more tomorrow; initial impressions are good.

storage/mountscan/mountscan.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+	return mounts, nil
+}
+
+// Utility functions
+func convertToUint(raw string) (uint, error) {
+	data, err := strconv.ParseUint(raw, 10, 64)
+	if err != nil {
+		return 0, errored.Errorf("Invalid data for converstion %q", raw).Combine(err)
```
erikh: "conversion"

storage/mountscan/mountscan.go
yuva29: I wanted this too be generic in `mountscan`. Also, this becomes handy when we have few more drivers down the line. Let me know your thoughts.

erikh: Seems good, my knee-jerk reaction was at the fact we don't need this for anybody but ceph, but the code itself is generic enough so that when something does come along we'll be good.

erikh:
LGTM, merging.

PR #391, opened 2016-08-07
Abstract config part #1

erikh:
Note that this is both unfinished and un-integrated. It is unit tested,
however.

This abstraction brings in a first-class home-built client that drives the
etcd client. It does this by abstracting away the details of both path
construction and client semantics.

Entities are the old notion of "config" objects that we used before.
Global, Policy are represented mostly in full now. Other objects will come
in a second patch.

This is the first part of a series of patches to support consul.

@mapuri if you have time I could really use your input here.

for our in-house committers, let's have a meeting to discuss this patch before discussing it on the tracker.


client/client.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+
+	"golang.org/x/net/context"
+)
+
+// ExistType is used to indicate whether or not a value can be operated on due
+// to the state of the existing node.
+type ExistType uint
+
+var toInitialize = []EntityCollection{}
```
dseevr: Is this for future use?  I don't see any other references to it.

client/client.go
```diff
@@ -0,0 +1,100 @@
...
+	Dump(string) (string, error)
+	Path() *Pather
+}
+
+// GetOptions are the retrieve options our abstractions support. These
+// properties will be translated on a per-driver basis to the appropriate
+// idioms.
+type GetOptions struct {
+	Recursive bool
+	Sort      bool // FIXME consul might not have this; is it necessary for anything?
```
dseevr: I know `ListPolicyRevisions()` (used in `volcli policy history list $name`) relies on this currently.  I see a few other "Sort: true" instances in the code, as well.  If we decide to drop support for sorted results from the datastore, we should add an issue to manually sort the responses in those spots.

erikh: Yeah, we'll have to implement separately in the consul client if necessary.

client/client.go
erikh: Actually this needs to be pruned, thanks.

client/client.go
dseevr: Ahh, good idea.  We can just reimplement it for stores which lack it as a feature.

client/entities/global.go
```diff
@@ -0,0 +1,198 @@
...
+	return content, nil
+}
+
+// Published returns a copy of the current global with the parameters adjusted
+// to fit the published representation. To see the internal/system/canonical
+// version, please see Canonical() below.
+//
+// It is very important that you do not run this function multiple times
+// against the same data set. It will adjust the parameters twice.
+func (g *Global) Published() *Global {
```
dseevr: What would be an example usage of this function?  I don't see it being called anywhere internally

erikh: oh this is a direct port of the config lib. the API server uses this to present to the user the external (duration-divided) values for re-uploading; the other `Canonical()` method does the inverse.

client/client.go
```diff
@@ -0,0 +1,100 @@
...
+	Value     []byte
+}
+
+// EntityClient is a client that is capable of performing events for a given
+// class of Entity.
+type EntityClient interface {
+	Path() *Pather
+	Get(string) (Entity, error)
+	Watch(string) (chan interface{}, chan error, error)
+	WatchStop(string) error
```
dseevr: Two questions here:

1) It seems like PolicyClient doesn't actually implement this function.  Why is that?

2) Based on the implementation for GlobalClient, it seems like the error return value is unnecessary.  Is there a reason for it?

erikh: 1) because it's not finished code; sorry to waste your time on that.
2) I think this kind of stuff ("what is necessary and what is not") will become more obvious as the integration starts happening. Unless you see a functional design problem with it, let's leave it be for now.

yuva29:
I'm not sure If i'm thinking right. Please correct me if i'm wrong.

This is how I see the whole abstraction

- client {interface which provides the outline for any data store, defines the minimal set of funcs to be implemented in a any datastore}
- Etcd, Consul, etc. {Provides the minimal implementation + anything specific to the data store }
- It will be great if we abstract all Policy, Volume, Global as a single object in etcd/consul. https://github.com/erikh/volplugin/blob/5544e2cedc2728ee9e4048e535d27ec317e38236/client/entities/global.go#L130
- There should be only one implementation of client interface, we don't need PolicyClient, GlobalClient, etc.

sorry, I really don't get the significance of each client. As our object space grows, we need to keep adding the code to support them. If we abstract them, I feel the changes will be minimal.

erikh:
I don't understand the relevancy of the linked code, can you explain? If it's about the global's use of the API, the globals don't have to do anything special because there's only ever one of them stored in the DB with a static name.

yuva29:
Ah.. that's a bad link. https://github.com/erikh/volplugin/blob/5544e2cedc2728ee9e4048e535d27ec317e38236/client/entities/structs.go#L23-L60 
Guess this might help, we are defining a client for each of our object.

dseevr:
Overall changes LGTM

erikh:
right, I don't like that much either. If you have an idea of how you'd like to change it, I could really use someone else thinking about it too.

erikh:
The function of the entity clients is to bridge the entity representation of itself with the object store it needs to be written to.

perhaps this would be better? @dseevr @unclejack do you have opinions on this as well?

```go
func NewGlobal(client *client.Client) { }

func (g *Global) Publish(name string) {
  // json
  // push to client
}
```

dseevr:
I like that dependency injection style.

Even if we keep the current system, some of the duplication could be addressed by creating a base type and embedding it:

```go
type EntityClientBase struct {
	client client.Client
	prefix *client.Pather
}

// PolicyClient is a client to fetch policies.
type PolicyClient struct {
	EntityClientBase
}

// ... and so on
```

yuva29:
How about this ?

```
func (c *client)Publish(name string, value interface{}) {
...
}
```

erikh:
what would the implementation of this look like?

erikh:
@dseevr great point about the embed; I think that might work too with minimal effort, but I'll need to play.

https://golang.org/doc/effective_go.html#embedding for those who are not familiar.

it'll be interesting to weigh the two options later tonight and see which plays better with tests, etc. (they're my only real metric on the API ATM)

erikh:
the embeds just aren't going to map; they "save" code, but they're implementation specific so only kinda sorta. At least, if I'm missing something please let me know.

erikh:
I think I'm going to close this, I want to change the design dramatically and the comments will no longer be relevant here.

I'm going to explore something similar to yuva's API: a visitor w/ methods instead of just a struct though.

PR #393, opened 2016-08-09
api: implement docker API endpoint unit tests.

erikh:
I'll be adding these incrementally as I get time. PTAL

### DIFF
--- a/api/impl/docker/docker_test.go
+++ b/api/impl/docker/docker_test.go
@@ -1,0 +1,163 @@
+package docker
+
+import (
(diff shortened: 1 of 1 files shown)


dseevr:
Looks like the [build failed with a VERR_ALREADY_EXISTS](http://contiv.ngrok.io/job/Volplugin_CI/1024/console).

How is this happening?  We check for a previous disk image in both cases we create them in Vagrantfile

erikh:
@mapuri do you have time to look into this?

PR #394, opened 2016-08-09
Support for volplugin, volsupervisor, apiserver in containers.

vvb:
- Making volplugin, volsupervisor, apiserver systemd services.
    - volplugin.service which uses /usr/bin/volplugin.sh
    - volsupervisor.service which uses /usr/bin/volsupervisor.sh
    - apiserver.service which uses /usr/bin/apiserver.sh

- volplugin.sh, volsupervisor.sh, apiserver.sh use
        /usr/bin/contiv-vol-run.sh to start respective containers.

/usr/bin/contiv-vol-run.sh has all the relevant docker run commands

- Changes to make sure that the system-tests use these containers instead.
- Verified that all unit-tests and system-tests are working


build/scripts/contiv-vol-run.sh
```diff
@@ -0,0 +1,60 @@
...
+set -e
+
+# cleanup the older container
+docker rm -f "$1" &>/dev/null || :
+
+## test for shared mount capability
+if ! grep "MountFlags" /lib/systemd/system/docker.service | grep shared &>/dev/null
+then
+  echo 1>&2 "Docker cannot run volplugin in its current state."
+  echo 1>&2 "You must change docker's systemd MountFlags to equal 'shared'"
```
mapuri: just curious, do you know why is this required?

I have been trying out kubernetes bring up in containers recently (using hyperkube) and looks like it requires MountFlags to be set to slave (https://github.com/kubernetes/kubernetes/issues/4869#issuecomment-194149162). So, we are kind of conflicting with that requirement.

erikh: yes, because of how slave and shared mounts work. Every volume driver needs this pretty much.

erikh: I don't understand, this is the exact thing we prescribe in our docs too. Is there some reason this can't be done in hyperkube? Maybe I missed that.

mapuri:
@vvb 

LGTM. Pretty clean and good quality work! However, I am wondering once we move this to ansible many of the script files will need to removed. Do you see any benefit of bringing this change through ansible instead in first place?

build/scripts/contiv-vol-run.sh
mapuri: > Is there some reason this can't be done in hyperkube?

Yeah, I think hyperkube requires some host volumes (like /var/lib/kubelet) to be mounted in shared mode, which fails if Docker is started with `MountFLags=slave`. I think moer details are there in the issue thread I shared above.

mapuri: I was just wondering if the requirement for volplugin is less strict then may be we can address it ourselves. I will look more into hyperkube as well, I just started playing with it.

vvb:
@mapuri yes, most of the files that are added in build/scripts/* will move to ansible/roles/contiv_storage/files or templates.. doing it here first was more about getting things to work - all tests to pass. 
@erikh and I also agreed that this should be done in ansible. 
This PR is more for code and structure review right now.

mapuri:
@vvb sounds good, thanks for clarifying !

Makefile
```diff
@@ -73,39 +73,56 @@ unit-test-nocoverage-host:
...
 	sudo systemctl start etcd
 
-run-volplugin: run-etcd
-	sudo pkill volplugin || exit 0
-	sudo -E nohup bash -c '$(GUESTBINPATH)/volplugin &>/tmp/volplugin.log &'
+docker-image:
+	docker build -t contiv/volplugin .
+
+create-systemd-services:
+	sudo cp '${GUESTGOPATH}/build/scripts/volplugin.service' /etc/systemd/system/
```
erikh: I'll be pointing out things that need to go to ansible here; this is one of them.

build/scripts/apiserver.sh
```diff
@@ -0,0 +1,33 @@
...
+    # now just sleep to keep the service up
+    mkfifo "/tmp/apiserver-fifo"
+    < "/tmp/apiserver-fifo"
+    ;;
+
+stop)
+    echo stopping apiserver
+    rm -f /tmp/apiserver-fifo
+    docker stop apiserver
+    docker rm -v apiserver
```
erikh: why are we removing volumes here? this doesn't remove the container I don't think.

build/scripts/contiv-vol-run.sh
erikh: Right. This is how mount propagation works.

more info here: https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt

erikh: in summary, the mounts will not propagate to the host bind-mount unless the shared flag is set on the parent mount point.

build/scripts/volsupervisor.sh
```diff
@@ -0,0 +1,34 @@
...
+start)
+
+    rm -f /tmp/volsupervisor-fifo
+
+    set -e
+    echo starting volsupervisor
+    /usr/bin/contiv-vol-run.sh volsupervisor
+
+    # now just sleep to keep the service up
+    mkfifo "/tmp/volsupervisor-fifo"
```
erikh: just use `read foo` here. it's cheaper and less confusing.

systemtests/init_test.go
```diff
@@ -48,9 +49,12 @@ func (s *systemtestSuite) SetUpSuite(c *C) {
 	if err != nil {
 		iter = defaultIterations
 	}
+	time.Sleep(time.Second)
```
erikh: these sleeps aren't necessary

systemtests/util_test.go
```diff
@@ -417,39 +417,40 @@ func restartNetplugin(node remotessh.TestbedNode) error {
 
 func startVolsupervisor(node remotessh.TestbedNode) error {
 	log.Infof("Starting the volsupervisor on %q", node.GetName())
-	return node.RunCommandBackground("(sudo -E nohup `which volsupervisor` </dev/null 2>&1 | sudo tee -a /tmp/volsupervisor.log) &")
+	defer time.Sleep(20 * time.Second)
```
erikh: 20 seconds is way too long. Remove this sleep and rely on the wait commands.

systemtests/util_test.go
```diff
@@ -417,39 +417,40 @@ func restartNetplugin(node remotessh.TestbedNode) error {
...
+	return node.RunCommand("sudo systemctl stop volsupervisor")
 }
 
 func startAPIServer(node remotessh.TestbedNode) error {
 	log.Infof("Starting the apiserver on %q", node.GetName())
-	err := node.RunCommandBackground("(sudo -E nohup `which apiserver` </dev/null 2>&1 | sudo tee -a /tmp/apiserver.log) &")
+	err := node.RunCommandBackground("sudo systemctl start apiserver")
 	log.Infof("Waiting for apiserver startup on %q", node.GetName())
-	time.Sleep(time.Second)
+	time.Sleep(30 * time.Second)
```
erikh: also too long.

systemtests/util_test.go
```diff
@@ -417,39 +417,40 @@ func restartNetplugin(node remotessh.TestbedNode) error {
...
 }
 
 func startVolplugin(node remotessh.TestbedNode) error {
 	log.Infof("Starting the volplugin on %q", node.GetName())
-	defer time.Sleep(10 * time.Millisecond)
-
-	// FIXME this is hardcoded because it's simpler. If we move to
-	// multimaster or change the monitor subnet, we will have issues.
-	return node.RunCommandBackground("(sudo -E `which volplugin` </dev/null 2>&1 | sudo tee -a /tmp/volplugin.log) &")
+	defer time.Sleep(20 * time.Second)
```
erikh: see also

systemtests/volplugin_test.go
```diff
@@ -102,6 +102,7 @@ func (s *systemtestSuite) TestVolpluginCrashRestart(c *C) {
 	time.Sleep(time.Second)
 	c.Assert(startVolplugin(s.vagrant.GetNode("mon0")), IsNil)
 	c.Assert(waitForVolplugin(s.vagrant.GetNode("mon0")), IsNil)
+	time.Sleep(30 * time.Second)
```
erikh: this should not be necessary, but I know what's probably causing this; let's look at why it is later and see how we can reduce this if possible.

erikh:
Be careful on adding sleeps to the systemtests, specifically in heavily repeated areas (such as restarting volplugin). It really adds up and our tests are already slow.

erikh:
additionally, if we're adding sleeps to the system tests I worry we're papering over bugs that shouldn't exist.

vvb:
sure, will look at reducing/removing sleeps

build/scripts/contiv-vol-run.sh
mapuri: thanks @erikh 

Actually, I was mistaken. Reading further even hyperkube requires `Mountflags` to be set to `shared`, so we seem to be good for now. Sorry for the distraction, but I ended up learning something here !

build/scripts/apiserver.sh
vvb: `docker rm -v` will remove the container and any volumes associated with it.

systemtests/util_test.go
```diff
@@ -392,7 +392,7 @@ func waitForAPIServer(node remotessh.TestbedNode) error {
 
 func waitForVolplugin(node remotessh.TestbedNode) error {
 	log.Infof("Checking if volplugin is running on %q", node.GetName())
-	err := runCommandUntilNoError(node, "pgrep -c volplugin", 10)
+	err := runCommandUntilNoError(node, "docker ps | grep volplugin | grep -v volsupervisor | grep -v apiserver", 10)
```
vvb: `docker inspect volplugin` does not really guarantee that the container is `running`.
Shall I use the below instead,
`docker inspect -f {{.State.Running}} volsupervisor | grep true` and check the exit code.

erikh: sgtm

build/scripts/apiserver.sh
erikh: I always thought it just removed volumes.... huh

erikh:
This looks very good; please do not take my comments otherwise. Just needs some touchups.

build/scripts/build-volplugin-containers.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+docker build -t contiv/volplugin .
+
+# Ensure that docker is running with MountFlags=shared
+sudo sed -i 's/MountFlags=slave/MountFlags=shared/g' /usr/lib/systemd/system/docker.service
+sleep 1
```
dseevr: Is this 1 second sleep necessary?  `systemctl daemon-reload` won't run before `sed` is finished, so I don't think there's any possible race condition here...

dseevr:
LGTM aside from what was already noted

vvb:
thanks @erikh @mapuri @dseevr for the comments. making the changes accordingly. will update the PR.

build/scripts/contiv-vol-run.sh
erikh: no distraction at all! useful context in these tickets to have.

erikh:
LGTM, but commit messages/squashing need work. Cannot merge until this is complete.

The ansible will come in the next set of patches, but for the interim let's merge this as the system test changes are likely going to require rebase over rebase over rebase as time presses on.

Squash your commits and prefix them with the directory you affect, e.g.:

```
Dockerfile,Makefile,build,systemtests: support running test suite in containers
```

You'll notice plenty of other examples in our commit messages on master. Let me know when this is done and i'll hit the merge button. Thanks!

erikh:
Congrats!

You have no idea how much I appreciate this. I was working on this for months. Thank you, @vvb.

PR #400, opened 2016-08-11
build: restart stopped containers instead of new ones

vvb:
`systemctl stop volplugin/volsupervisor/apiserver` now stops the container and does not remove it. when a `systemctl start` is issued, the same container is started again. a `rm` target is also added to `volplugin.sh/volsupervisor.sh/apiserver.sh` which will optionally remove the containers too.


erikh:
LGTM. Pushing this through via executive veto...

PR #401, opened 2016-08-11
volmigrate, Makefile: add volmigrate tool which can list and run schema migrations

### LINKED ISSUE #353: etcd migrations system
Research if there's something like this first, but to appropriately convert databases between versions we should ensure that we have a system like this in place. It will become more critical as more people start using the product.

dseevr:
Commands:

`volmigrate list`          -> see all migration versions and descriptions
`volmigrate run [version]` -> run all pending migrations up to [version]
`volmigrate run`           -> run all pending migrations
`volmigrate version`       -> see the local schema version (0 if no migrations have ever run)

The current schema version is held in the "/volplugin/schema-version" key.  Each migration will increment the version by 1 (i.e., the first migration is 1, the second is 2, and so on).  Migration numbers can't be reused.

There is an optional global --silent flag to suppress the confirmation prompt when attempting to run migrations.

Signed-off-by: Bill Robinson <dseevr@users.noreply.github.com>

Fixes #353

### DIFF
--- a/volmigrate/backend/backend.go
+++ b/volmigrate/backend/backend.go
@@ -1,0 +1,38 @@
+package backend
+
+const SchemaVersionKey = "schema-version"
+
+// Backend defines the interface for a type which implements the functionality to talk
+// to a given datastore (etcd2, etcd3, consul, etc.)
+type Backend interface {
+	// CurrentSchemaVersion returns the version of the last migration which was successfuly run.
+	// This is used to ensure that the migration we're about to run hasn't been run already.
+	// If the version key does not exist, it will return 0.
+	CurrentSchemaVersion() int64
+
+	// CreateDirectory creates a directory at the target path.
+	// It acts like mkdir -p, e.g., it won't fail if the directory already exists and
+	// can create any depth of nested folder hierarchy.
+	// If the target path exists and is a file, it will return an error.
+	CreateDirectory(path string) error
+
+	// CreateKey will create a key at the target location.
+	// If the target key already exists or is a directory, it will return an error.
+	CreateKey(path string, contents []byte) error
+
+	// DeleteDirectory will recursively delete a target directory.
+	// If the target directory doesn't exist, it's not considered a failure.
+	// If the target directory is a file, it will return an error.
+	DeleteDirectory(path string) error
+
+	// DeleteKey will delete the target key if it exists.
+	// If the target key doesn't exist, it's not considered a failure.
+	// If the target key is a directory, it will refuse to remove it and will return an error.
+	DeleteKey(path string) error
+
+	// Name returns the name of the engine ("etcd2", "etcd3", "consul", etc.)
+	Name() string
+
+	// UpdateSchemaVersion records the version number of the latest migration which successfully ran.
+	UpdateSchemaVersion(version int64) error
+}
--- a/volmigrate/backend/etcd2/etcd2.go
+++ b/volmigrate/backend/etcd2/etcd2.go
@@ -1,0 +1,125 @@
+package etcd2
+
+import (
+	"fmt"
+	"log"
+	"path"
+	"strconv"
+
+	"github.com/contiv/errored"
+	"github.com/contiv/volplugin/errors"
+	"github.com/contiv/volplugin/volmigrate/backend"
+	"github.com/coreos/etcd/client"
+	"golang.org/x/net/context"
+)
+
+func New(prefix string, etcdHosts []string) *Engine {
+	etcdCfg := client.Config{
+		Endpoints: etcdHosts,
+	}
+
+	etcdClient, err := client.New(etcdCfg)
+	if err != nil {
+		log.Fatalf("Failed to create etcd client: %s", err)
+	}
+
+	return &Engine{
+		etcdClient: client.NewKeysAPI(etcdClient),
+		prefix:     prefix,
+	}
+}
+
+// Engine is a migration engine for an etcd v2 datastore.
+type Engine struct {
+	etcdClient client.KeysAPI
+	prefix     string
+}
+
+// CurrentSchemaVersion returns the version of the last migration which was successfuly run.
+// If no previous migrations have been run, the schema version is 0.
+// If there's any error besides "key doesn't exist", execution is aborted.
+func (e *Engine) CurrentSchemaVersion() int64 {
+	resp, err := e.etcdClient.Get(context.Background(), path.Join(e.prefix, backend.SchemaVersionKey), nil)
+	if err != nil {
+		if err := errors.EtcdToErrored(err); err != nil && err == errors.NotExists {
+			return 0 // no key = schema version 0
+		}
+
+		log.Fatalf("Unexpected error when looking up schema version: %v\n", err)
+	}
+
+	i, err := strconv.Atoi(resp.Node.Value)
+	if err != nil {
+		log.Fatalf("Got back unexpected schema version data: %v\n", resp.Node.Value)
+	}
+
+	return int64(i)
+}
+
+// CreateDirectory creates a directory at the target path.
+func (e *Engine) CreateDirectory(target string) error {
+	fmt.Println("Creating directory: " + target)
+	if _, err := e.etcdClient.Set(context.Background(), path.Join(e.prefix, target), "", &client.SetOptions{Dir: true}); err != nil {
+		return errored.Errorf("Failed to create directory: %s", err)
+	}
+
+	return nil
+}
+
+// CreateKey will create a key at the target location.
+func (e *Engine) CreateKey(target string, contents []byte) error {
+	fmt.Println("Creating key: " + target)
+	if _, err := e.etcdClient.Set(context.Background(), path.Join(e.prefix, target), string(contents), nil); err != nil {
+		errored.Errorf("Failed to create key: %s", err)
+	}
+
+	return nil
+}
+
+// DeleteDirectory will recursively delete a target directory.
+func (e *Engine) DeleteDirectory(target string) error {
+	fmt.Println("Deleting directory: " + target)
+
+	if _, err := e.etcdClient.Delete(context.Background(), path.Join(e.prefix, target), &client.DeleteOptions{Dir: true, Recursive: true}); err != nil {
+		return errored.Errorf("Failed to delete directory: %s", err)
+	}
+
+	return nil
+}
+
+// DeleteKey will delete the target key if it exists.
+func (e *Engine) DeleteKey(target string) error {
+	fmt.Println("Deleting key: " + target)
+
+	if _, err := e.etcdClient.Delete(context.Background(), path.Join(e.prefix, target), &client.DeleteOptions{Dir: false}); err != nil {
+		return errored.Errorf("Failed to delete key: %s", err)
+	}
+
+	return nil
+}
+
+// Name returns the name of the engine ("etcd2", "etcd3", "consul", etc.)
+func (e *Engine) Name() string {
+	return "etcd2"
+}
+
+// UpdateSchemaVersion records the version number of the latest migration which successfully ran.
+// If the new version number is <= the current version number, it will log a fatal error.
+func (e *Engine) UpdateSchemaVersion(newVersion int64) error {
+	fmt.Printf("Updating schema version key to: %d\n", newVersion)
+
+	currentVersion := e.CurrentSchemaVersion()
+
+	// sanity check to make sure the version number is actually increasing
+	if newVersion <= currentVersion {
+		log.Fatalf("Cowardly refusing to update schema version to a version <= the current version.  Current: %d, Desired: %d\n", currentVersion, newVersion)
+	}
+
+	data := strconv.Itoa(int(newVersion))
+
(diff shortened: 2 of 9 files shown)


dseevr:
Note: this code isn't 100% done yet (there's some TODOs and no tests), but I want to get some feedback on the design/assumptions.

@erikh, @yuva29, @unclejack PTAL

erikh:
I'll get to this tonight. Awesome! Looking forward to this!

volmigrate/backend/etcd2/etcd2.go
```diff
@@ -0,0 +1,112 @@
...
+type Engine struct {
+	etcdClient client.KeysAPI
+	prefix     string
+}
+
+// CurrentSchemaVersion returns the version of the last migration which was successfuly run.
+func (e *Engine) CurrentSchemaVersion() int64 {
+	resp, err := e.etcdClient.Get(context.Background(), path.Join(e.prefix, backend.SchemaVersionKey), nil)
+	if err != nil {
+		// TODO: check that this is actually a Not Found
```
erikh: please see `errors.EtcdToErrored` which can help with this

volmigrate/migration.go
```diff
@@ -0,0 +1,41 @@
...
+// Migration represents a migration that volmigrate can run.
+type Migration struct {
+	Version     int64  // each new migration increments this value, must be >= 1
+	Description string // e.g., "volplugin 1.3 release"
+	runner      func(backend.Backend) error
+}
+
+// Run executes the migration using a supplied backend.Backend
+func (m *Migration) Run(backend backend.Backend) error {
+	log.WithFields(log.Fields{
```
erikh: specific reason for withfields?

erikh:
This looks great man. Nice work.

One suggestion: since `volmigrate` will be a static bin with the migrations pre-baked, can we also output in `volmigrate version` the version of the latest migration we support? More a UX thing than anything else.

volmigrate/migration.go
dseevr: None, I was just playing with the log library 👯

dseevr:
What kind of output are you thinking?  Something like this?

```
Current local schema version: 4
Newest available schema version: 5
```

Not sure what wording makes the most sense here.

volmigrate/backend/etcd2/etcd2.go
dseevr: Perfect, I assumed there was a way to do this so I didn't bother writing my own.  Thanks!

volmigrate/migration.go
erikh: ah ok. Let's leave this out then; I want to do more structured logging but would like to also keep things consistent. Once we get the chance we will.

erikh:
That communicates well enough, let's go with that.

volmigrate/backend/etcd2/etcd2.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return int64(i)
+}
+
+// CreateDirectory creates a directory at the target path.
+func (e *Engine) CreateDirectory(target string) error {
+	fmt.Println("Creating directory: " + target)
+	if _, err := e.etcdClient.Set(context.Background(), path.Join(e.prefix, target), "", &client.SetOptions{Dir: true}); err != nil {
+		return errored.Errorf("Failed to create directory: %s", err)
```
yuva29: @dseevr Do you think its better to create "ErrMigrationFailed" error and wrap all the errors along with it?

volmigrate/backend/etcd2/etcd2.go
```diff
@@ -0,0 +1,117 @@
...
+func (e *Engine) Name() string {
+	return "etcd2"
+}
+
+// UpdateSchemaVersion records the version number of the latest migration which successfully ran.
+func (e *Engine) UpdateSchemaVersion(version int64) error {
+	fmt.Printf("Updating schema version key to: %d\n", version)
+
+	data := strconv.Itoa(int(version))
+
```
yuva29: Do we need to check version>currentSchemaVersion() here? I'm not sure if its really required.

volmigrate/migration.go
```diff
@@ -0,0 +1,39 @@
...
+// Run executes the migration using a supplied backend.Backend
+func (m *Migration) Run(b backend.Backend) error {
+	fmt.Printf("Running migration #%d against %s backend\n", m.Version, b.Name())
+
+	if err := m.runner(b); err != nil {
+		return errored.Errorf("Encountered error during migration %d", m.Version).Combine(err)
+	}
+
+	if err := b.UpdateSchemaVersion(m.Version); err != nil {
+		return errored.Errorf("Successfully applied migration but failed to update schema version key").Combine(err)
```
yuva29: Apparently the exact version is not recorded here but the migration applied is still active. Is it the right behavior?

volmigrate/backend/backend.go
```diff
@@ -0,0 +1,38 @@
+package backend
+
+const SchemaVersionKey = "schema-version"
+
+// Backend defines the interface for a type which implements the functionality to talk
+// to a given datastore (etcd2, etcd3, consul, etc.)
+type Backend interface {
```
yuva29: I totally like the comments here! very helpful 👍

volmigrate/migration.go
dseevr: Without transaction support, there's nothing we can do except print a warning and have the end user rectify the situation manually.

In an optimal world, the entire migration + the schema version key update would be a single atomic unit of work.  My original design did include `BeginTransaction()` and `CommitTransaction()` as part of the Backend interface, but since etcd2 doesn't support transactions, they were just a NOOP.  If we want to support transactions in the future, it's trivial to add them (add interface functions + wrap `m.Run()` in migration.go).

volmigrate/schema_migrations.go
```diff
@@ -0,0 +1,62 @@
...
+
+func init() {
+	usedVersionNumbers = make(map[int64]bool)
+	registerMigrations()
+	sort.Sort(Migrations(availableMigrations))
+}
+
+// registerMigration makes volmigrate aware of a new migration that can be run.
+// If the version number has already been used, it will log a fatal error.
+func registerMigration(version int64, description string, f func(backend backend.Backend) error) {
```
yuva29: You can accept uint here to avoid L24 check

volmigrate/backend/etcd2/etcd2.go
dseevr: This function is only called from one spot (`Migration`'s `Run()` function), so it can only ever execute if a migration was deemed to be necessary, and necessary implies `m.version > currentVersion`.  I can add this as a sanity, though.  Good catch.

yuva29:
@dseevr And, I think it will be better if we can record the time at which the migration ran. 
Also, we can add a feature to provide the list of migrations ran and at what times
WDYT @erikh

volmigrate/schema_migrations.go
dseevr: uint would still allow 0, so I'd still need to check anyways.  0 is the schema version of an etcd keyspace which has never run any migrations.

volmigrate/volmigrate.go
```diff
@@ -0,0 +1,170 @@
...
+	promptBeforeRunning(ctx, "You have requested to run all pending migrations.")
+	return runAllMigrations(e)
+}
+
+func runAllMigrations(b backend.Backend) (bool, error) {
+	return runMigrationsUpTo(b, latestMigrationVersion())
+}
+
+func findMigrationByVersion(cfg *config.Client, version int64) (*Migration, bool) {
+	for _, m := range availableMigrations {
```
yuva29: Isn't it the below true here?

```go
if i > len(availableMigrations) {
return nil, false
} else {
return availableMigrations[m-1], true
}
```

volmigrate/backend/etcd2/etcd2.go
dseevr: By "wrap" do you mean use errored.Combine() to encapsulate the actual error inside the ErrMigrationFailed error?

There's no recovering from a failure during a transaction unless you explicitly ignore the error in your code.  I'm not sure what value having an error type which just says "migration failed" adds over the current setup.  What are you thinking?

volmigrate/volmigrate.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return nil, false
+}
+
+func runMigrationsUpTo(b backend.Backend, targetVersion int64) (bool, error) {
+	migrationsRun := 0
+
+	currentVersion := b.CurrentSchemaVersion()
+
+	for _, m := range availableMigrations {
```
yuva29: Don't we need to run the migrations in order 1.. targetVersion ?

dseevr:
That's a good point.  Rails does record the UNIX timestamp of when a migration was run.  It does this by creating a dedicated table called `schema_revisions` where it inserts one record per migration that has run.

If we decide to add this, we could change the keyspace like:

```
/volplugin/schema/current -> this has the id of the last migration that was run
/volplugin/schema/revisions/1/applied-at -> unix timestamp
/volplugin/schema/revision/2/applied-at -> unix timestamp
...
etc.
```

volmigrate/volmigrate.go
dseevr: That's what this currently does, unless you've spotted a bug!

volmigrate/schema_migrations.go
yuva29: You're right!

volmigrate/volmigrate.go
dseevr: Good point, but that also assumes that there are no gaps in the migration version sequence.   Currently, `registerMigration()` *would* allow that behavior.

The reason I made the version ID explicit (i.e., an argument to `registerMigration()`) is because having it auto-increment when a new migration is registered makes migration IDs position-dependent in the source.  It would also make it harder to find a specific migration by ID when looking at the source code.

As a middle ground, I can ensure that when a migration is registered it has the expected ID (the next in sequence) and abort otherwise.  Then this simpler code you've written would work and there's no danger of position-dependent migration IDs or sequence gaps.

Thanks, I will do this.

volmigrate/backend/etcd2/etcd2.go
yuva29: hmm.. Got it. I would like to know what @erikh thinks about it?

volmigrate/volmigrate.go
dseevr: Based on the other comment I just left, once I enforce that the next migration being registered has the expected sequential ID, this is guaranteed to be in order.  Right now, it's manually sorted in `volmigrate.init()`.  That and the `Migrations` type can be deleted once I make this change.

yuva29:
Before getting into the comments again, Overall the code looks good and special mention to the comments 👍

dseevr:
@yuva29 Based on your comments, I was able to replace the array of migrations with a map, get rid of a few O(n) traversals, and simplify a bunch of code.  There's now a guarantee that there are no gaps in the migration sequence (i.e., they MUST be 1,2,3,4... etc.)

I also added a sanity that you can't downgrade the schema version.  :smile 

Thanks for the feedback!  Time to add tests...

volmigrate/backend/etcd2/etcd2.go
erikh: Hmm. I like the error type, but if we're not going to use it, that's fine. I would be extra careful to clearly describe what is happening (namely about migrations happening), though.

volmigrate/migration.go
erikh: I think this is OK for now.

I think we should also start another ticket with "nice to have's" about this feature, which will allow this to get through yet still see some of these things happen.

erikh:
Ok, don't sweat doing anything more to this feature-wise. I think we're good for now. Please put in another ticket yuva or bill on the features you'd like to see soon. That way they can get knocked out when we have free moments.

volmigrate/backend/etcd2/etcd2.go
erikh: We know this is the volmigrate tool so I'm not sure this is really as descriptive as it sounds, on a second pass. No need for the type imo but I'm not going ot fight either way.

volmigrate/schema_migrations.go
```diff
@@ -0,0 +1,64 @@
...
+// If a migration returns an error, volmigrate execution is aborted and the schema version key
+// is not updated.  If an operation in a migration can legitimately fail without causing problems,
+// don't check the error value from the Create/Delete call.
+// Paths are relative to /volplugin and don't require a slash at the beginning:
+// e.g., CreateDirectory("foo/bar") will create "/volplugin/foo/bar"
+func registerMigrations() {
+
+	// This migration does nothing, but the backend.SchemaVersionKey key will be
+	// created and populated after it finishes.
+	registerMigration(1, "Create initial schema version key", func(b backend.Backend) error {
```
erikh: one nit. can we put these functions in their own package or file so that they can be easily edited by many people? best to start the pattern now than to clean it up later, I think. It doesn't even have to be its own package, just another file e..g, `migrations.go`.

erikh:
ok, LGTM.

erikh:
ok -- @yuva29 @unclejack LGTM?

PR #402, opened 2016-08-12
Fix dangling ttl goroutine

erikh:
This fixes an ordering issue in the locking code where a goroutine would dangle, or be left unchecked, refreshing a TTL on a lock that no longer needed to be held. Additionally, if the mount failed the lock would not be cleared.

I also fixed a small test issue with taking enough time to boot the API server.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -185,6 +185,33 @@ func (a *API) List(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+type mountState struct {
+	w          http.ResponseWriter
+	err        error
+	ut         *config.UseMount
+	driver     storage.MountDriver
+	driverOpts storage.DriverOptions
+	volConfig  *config.Volume
+}
+
+// triggered on any failure during call into mount.
+func (a *API) clearMount(ms mountState) {
+	log.Errorf("MOUNT FAILURE: %v", ms.err)
+
+	if err := ms.driver.Unmount(ms.driverOpts); err != nil {
+		// literally can't do anything about this situation. Log.
+		log.Errorf("Failure during unmount after failed mount: %v %v", err, ms.err)
+	}
+
+	if err := a.Lock.ClearLock(ms.ut, (*a.Global).Timeout); err != nil {
+		a.HTTPError(ms.w, errors.RefreshMount.Combine(errored.New(ms.volConfig.String())).Combine(err).Combine(ms.err))
+		return
+	}
+
+	a.HTTPError(ms.w, errors.MountFailed.Combine(ms.err))
+	return
+}
+
 // Mount is the request to mount a volume.
 func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
 	request, err := a.ReadMount(r)
@@ -203,13 +230,13 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	volName := volConfig.String()
-	if !volConfig.Unlocked {
-		ut := &config.UseMount{
-			Volume:   volName,
-			Reason:   lock.ReasonMount,
-			Hostname: a.Hostname,
-		}
+	ut := &config.UseMount{
+		Volume:   volName,
+		Reason:   lock.ReasonMount,
+		Hostname: a.Hostname,
+	}
 
+	if !volConfig.Unlocked {
 		// XXX the only times a use lock cannot be acquired when there are no
 		// previous mounts, is when in locked mode and a mount is held on another
 		// host. So we take an indefinite lock HERE while we calculate whether or not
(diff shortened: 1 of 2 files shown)


api/handlers.go
```diff
@@ -239,22 +239,27 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
...
 		a.HTTPError(w, errors.MountFailed.Combine(err))
 		return
 	}
 
 	a.MountCollection.Add(mc)
 
+	// Only perform the TTL refresh if the driver is in unlocked mode.
+	if !volConfig.Unlocked {
+		if err := a.startTTLRefresh(volName); err != nil {
+			a.HTTPError(w, err)
```
yuva29: @erikh Do we ever get into this situation here? If it does, the published lock dangles again right ?

erikh: hmm yeah.

dseevr:
LGTM

erikh:
Can I get @yuva29 @vvb or @unclejack to review this?

api/handlers.go
```diff
@@ -239,35 +239,62 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
...
 
+	// Only perform the TTL refresh if the driver is in unlocked mode.
+	if !volConfig.Unlocked {
+		if err := a.startTTLRefresh(volName); err != nil {
+			a.RemoveStopChan(volName)
+
+			if err := a.clearLock(ut); err != nil {
+				a.HTTPError(w, err)
+				return
+			}
```
yuva29: @erikh I don't know what's the right action to be performed here. but, this change will make the mount at L242 untracked. I suspect it will create issues while snapshotting locked volumes which purely relies on mount locks.

erikh: explain in detail, please.

erikh: the refresh is killed by `RemoveStopChan` and the permanent lock is cleared by `ClearLock()`. I don't think what you're suggesting is possible so please spell it out for me.

yuva29: right. But still the volume is mounted right?

erikh: damnit.

erikh:
@yuva29 @dseevr @unclejack PTAL. I'm not sure I like the struct, but I don't really want to introduce anything larger and the function prototype was crazy.

api/handlers.go
```diff
@@ -185,6 +185,33 @@ func (a *API) List(w http.ResponseWriter, r *http.Request) {
...
+// triggered on any failure during call into mount.
+func (a *API) clearMount(ms mountState) {
+	log.Errorf("MOUNT FAILURE: %v", ms.err)
+
+	if err := ms.driver.Unmount(ms.driverOpts); err != nil {
+		// literally can't do anything about this situation. Log.
+		log.Errorf("Failure during unmount after failed mount: %v %v", err, ms.err)
+	}
+
+	if err := a.Lock.ClearLock(ms.ut, (*a.Global).Timeout); err != nil {
```
yuva29: I guess here you can call `clearLock(ut config.UseLocker)`, this will eliminate `volConfig` from the struct. And, you can return `err` and wrap it with `HTTPError` in the caller `Mount`(this will remove w, err from the struct). Hope it will make this function a little cleaner.

api/handlers.go
```diff
@@ -239,35 +266,52 @@ func (a *API) Mount(w http.ResponseWriter, r *http.Request) {
...
+		a.clearMount(mountState{w, err, ut, driver, driverOpts, volConfig})
 		return
 	}
 
 	a.MountCollection.Add(mc)
 
+	// Only perform the TTL refresh if the driver is in unlocked mode.
+	if !volConfig.Unlocked {
+		if err := a.startTTLRefresh(volName); err != nil {
+			a.RemoveStopChan(volName)
```
yuva29: `a.RemoveStopChan(volName)` can be moved to `clearMount`. It doesn't cause any harm if the channel is not there.

api/handlers.go
erikh: no that's a ton of prep for each call. I don't want to do that.

api/handlers.go
erikh: I don't know if this is true. I need to research it. Let's just leave it alone for now.

erikh:
@yuva29 are you ok with this patch in its current form?

yuva29:
@erikh the logic seems good to me. I'm not sure about the structure. But I think we can merge for now and refactor it later.

erikh:
to be clear, I have never liked this code, it's just... well... necessary. So perhaps in the near future we can discuss how to approrpiately remodel this code. Until then though, we need something that works.

0.4 would be a good target for the bigger changes.

PR #403, opened 2016-08-12
build,systemtests: wait for apiserver to be reachable

vvb:
We were hitting apiserver before it was ready to serve requests.
This PR uses `connwait` to wait for apiserver to be ready.
we also keep track of time it takes to become accessible.

### DIFF
--- a/systemtests/util_test.go
+++ b/systemtests/util_test.go
@@ -385,7 +385,13 @@ func waitForAPIServer(node remotessh.TestbedNode) error {
 		log.Infof("APIServer is running on %q", node.GetName())
 	}
 
-	time.Sleep(2 * time.Second)
+	then := time.Now()
+	err = runCommandUntilNoError(node, "connwait 127.0.0.1:9005", 60)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


dseevr:
LGTM

systemtests/util_test.go
```diff
@@ -384,7 +384,13 @@ func waitForAPIServer(node remotessh.TestbedNode) error {
...
 	}
-	time.Sleep(time.Second)
+
+	then := time.Now()
+	err = runCommandUntilNoError(node, "connwait 127.0.0.1:9005", 60)
+	if err == nil {
+		log.Infof("Took %s for apiserver on %q to be accessible", time.Since(then), node.GetName())
+	}
+
 	return nil
```
erikh: return err here since it could be non-nil

erikh:
Let's put this in `build/scripts/deps.sh` instead of `checks.sh`, and make a new reference to `deps.sh` in our make targets.

Looks great otherwise

vvb:
@erikh comments are taken care of. PTAL

erikh:
LGTM

PR #404, opened 2016-08-12
storage/mountscan/mountscan.go: correct the tolerated number of field…

erikh:
…s in mountinfo parser.

The mountinfo parser was too strict and parsing an 11-entry set which exists on
the 4.x kernels Ubuntu uses, but not the 3.10 kernels CentOS does. I've
adjusted it to contain the 11th field but not parse it in case we decide that
this compatibility is no longer required.

### DIFF
--- a/storage/mountscan/mountscan.go
+++ b/storage/mountscan/mountscan.go
@@ -14,7 +14,7 @@ const (
 	mountInfoFile           = "/proc/self/mountinfo"
 	deviceInfoFile          = "/proc/devices"
 	nfsMajorID              = 0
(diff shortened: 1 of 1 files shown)


dseevr:
LGTM

erikh:
@yuva29 can you sign off on this if it's ok?

yuva29:
@erikh I assume the mapping of fields works well with centOS. Here https://github.com/erikh/volplugin/blob/a5c24031d85cb36187059a373d56453ef8f9a29d/storage/mountscan/mountscan.go#L86-L94 If its fine, then changes LGTM

PR #406, opened 2016-08-12
volsupervisor, config, systemtests: Snapshot only volumes in use

### LINKED ISSUE #396: Volsupervisor should only snapshot when the volume is in use
We currently create the snapshot either way, I don't think pruned snaps of the same volume every 10m (for example) when the data is at rest leads to very useful snapshot logs.

yuva29:
fixes #396 

### DIFF
--- a/config/volume.go
+++ b/config/volume.go
@@ -405,3 +405,17 @@ func (cfg *Volume) ToDriverOptions(timeout time.Duration) (storage.DriverOptions
 func (cfg *Volume) String() string {
 	return path.Join(cfg.PolicyName, cfg.VolumeName)
 }
+
+// IsVolumeInUse checks if the given volume is mounted in any container
+func (c *Client) IsVolumeInUse(cfg *Volume) (bool, error) {
+	if !cfg.Unlocked {
+		if _, err := c.etcdClient.Get(context.Background(), c.use(UseTypeMount, cfg.String()), nil); err != nil {
+			if etcdErr := errors.EtcdToErrored(err); etcdErr != nil {
+				return !(etcdErr == errors.NotExists), nil
+			}
+			return true, err
+		}
+	}
+	// XXX To simplify the behavior, always return true for unlocked volumes as there are no mount lock exists for them.
+	return true, nil
+}
--- a/systemtests/integrated_test.go
+++ b/systemtests/integrated_test.go
@@ -222,7 +222,7 @@ func (s *systemtestSuite) TestIntegratedVolumeSnapshotCopy(c *C) {
 	targetName := genRandomVolume()
 	targetName2 := genRandomVolume()
 
-	c.Assert(s.createVolume("mon0", fqVolName, nil), IsNil)
+	c.Assert(s.createVolume("mon0", fqVolName, map[string]string{"unlocked": "true"}), IsNil)
 
 	time.Sleep(4 * time.Second)
 
--- a/systemtests/volsupervisor_test.go
+++ b/systemtests/volsupervisor_test.go
@@ -1,11 +1,63 @@
 package systemtests
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	. "gopkg.in/check.v1"
 )
+
+func (s *systemtestSuite) TestVolsupervisorSnapLockedVolume(c *C) {
+	if !cephDriver() {
+		c.Skip("Only ceph supports snapshots")
+		return
+	}
+
+	_, err := s.uploadIntent("policy1", "snaplockedvol")
+	c.Assert(err, IsNil)
+
+	volName := genRandomVolume()
+	fqVolName := fqVolume("policy1", volName)
+	c.Assert(s.createVolume("mon0", fqVolName, nil), IsNil) // locked volume
+
+	// XXX provides some time for the snap creation code to execute (snap for every 2s)
+	time.Sleep(4 * time.Second)
+	prevCount := 0
+	for count := 0; count < 5; count++ {
+		out, err := s.vagrant.GetNode("mon0").RunCommandWithOutput("sudo rbd snap ls policy1." + volName)
+		c.Assert(err, IsNil)
+		snapCount := len(strings.Split(out, "\n")) // returns 1 for empty "out"
+		// XXX Below if-statement executes only during 1st iteration
+		if len(strings.TrimSpace(out)) == 0 {
+			snapCount = 0
+		}
+		c.Assert(snapCount == prevCount, Equals, true)
+
+		start := time.Now()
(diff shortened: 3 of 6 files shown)


config/volume.go
```diff
@@ -405,3 +405,17 @@ func (cfg *Volume) ToDriverOptions(timeout time.Duration) (storage.DriverOptions
 func (cfg *Volume) String() string {
 	return path.Join(cfg.PolicyName, cfg.VolumeName)
 }
+
+// IsVolumeInUse checks if the given volume is mounted in any container
+func (c *Client) IsVolumeInUse(cfg *Volume) (bool, error) {
+	if !cfg.Unlocked {
+		if _, err := c.etcdClient.Get(context.Background(), c.use(UseTypeMount, cfg.String()), nil); err != nil {
+			if cErr, ok := err.(client.Error); ok {
```
dseevr: I just did something similar, but my implementation is a little different:  https://github.com/contiv/volplugin/pull/401/commits/b6e5acb5bf31f5b83182336a8dc356842775b361#diff-3fb318b7d8ab9f4eb7fd5d17b99eb42aR44

@erikh: what is the preferred method here?

config/volume.go
```diff
@@ -405,3 +405,17 @@ func (cfg *Volume) ToDriverOptions(timeout time.Duration) (storage.DriverOptions
...
+				return !(cErr.Code == client.ErrorCodeKeyNotFound), nil
+			}
+			return true, err
+		}
+	}
+	// XXX To simplify the behavior, always return true for unlocked volumes as there are no mount lock exists for them.
```
dseevr: If you check if the volume is unlocked and return at the beginning, you can get rid of one level of indentation in most of the function.  Not sure what is stylistically best here.

systemtests/volsupervisor_test.go
```diff
@@ -1,12 +1,61 @@
...
+	}
+
+	_, err := s.uploadIntent("policy1", "snaplockedvol")
+	c.Assert(err, IsNil)
+
+	volName := genRandomVolume()
+	fqVolName := fqVolume("policy1", volName)
+	c.Assert(s.createVolume("mon0", fqVolName, nil), IsNil) // locked volume
+
+	time.Sleep(4 * time.Second)
```
dseevr: Is there a way to poll for the volume creation to be finished rather than sleeping?

yuva29: This sleep is to ensure that no snaps has been created when the volume is not mounted.

dseevr: How does sleeping prevent snapshots from being created?  Sorry, I'm not really familiar with this stuff yet.

yuva29: So, the volume(locked - only one mount possible on the host; when one container mounts it, no one else can mount it) is actually not in use until its mounted in some container. Here I have created the volume with this [snapunlocked.json](https://github.com/yuva29/volplugin/blob/2734854e9243d5cdeb6b0b27bb34690e67891a5f/systemtests/testdata/ceph/snaplockedvol.json) policy (snaps are taken every 2 seconds).

At L25, the sleep of 4s (gives time for 2 snap creation) but the volume is locked and not mounted(Not in use) so there should be NO snaps. That's what is checked in the first iteration of the loop (snapCount == prevCount == 0). Hope this helps. Please let me know if I made it confusing

config/volume.go
yuva29: Then we have handle the case when the volume is in use (extra 3 lines of code like below).

```go
if cfg.Unlocked {
  return true, nil
}

if _, err := c.etcdClient.Get(context.Background(), c.use(UseTypeMount, cfg.String()), nil); err != nil {
  if cErr, ok := err.(client.Error); ok {
    return !(cErr.Code == client.ErrorCodeKeyNotFound), nil
	}
	return true, err
} else { // Volume is in USE
  return true, nil
}	
```

dseevr: Ahh, I see.  Please disregard.  :)

systemtests/volsupervisor_test.go
dseevr: Thanks for the explanation!  I see you added a comment for it 👍

config/volume.go
erikh: guys use `errors.EtcdToErrored` here and the `Contains/Combine` methods.

systemtests/integrated_test.go
```diff
@@ -222,7 +222,7 @@ func (s *systemtestSuite) TestIntegratedVolumeSnapshotCopy(c *C) {
 	targetName := genRandomVolume()
 	targetName2 := genRandomVolume()
 
-	c.Assert(s.createVolume("mon0", fqVolName, nil), IsNil)
+	c.Assert(s.createVolume("mon0", fqVolName, map[string]string{"unlocked": "true"}), IsNil)
```
erikh: why did this change?

erikh: ohj I think I see now. it's to get the older behaving tests passing right?

erikh:
LGTM minus the one comment. Thanks for turning this around so quickly.

systemtests/integrated_test.go
yuva29: Yep.

yuva29:
@erikh @dseevr  PTAL https://github.com/contiv/volplugin/pull/406/commits/1a606482375fb0cf87dc91dd64b49496d787661f I made these changes to avoid noisy error logs

dseevr:
LGTM

config/volume.go
```diff
@@ -405,3 +405,17 @@ func (cfg *Volume) ToDriverOptions(timeout time.Duration) (storage.DriverOptions
 func (cfg *Volume) String() string {
 	return path.Join(cfg.PolicyName, cfg.VolumeName)
 }
+
+// IsVolumeInUse checks if the given volume is mounted in any container
+func (c *Client) IsVolumeInUse(cfg *Volume) (bool, error) {
+	if !cfg.Unlocked {
+		if _, err := c.etcdClient.Get(context.Background(), c.use(UseTypeMount, cfg.String()), nil); err != nil {
+			if etcdErr := errors.EtcdToErrored(err); etcdErr != nil {
```
erikh: wrap your errors with `errors.EtcdToErrored`

erikh:
Please write a test to ensure that locked images that are not mounted are not snapped. Thanks!

yuva29:
@erikh [SnapLockedVol](https://github.com/yuva29/volplugin/blob/1a606482375fb0cf87dc91dd64b49496d787661f/systemtests/volsupervisor_test.go#L12) functions handles that case. And it also checks the behavior when its mounted and unmounted.

erikh:
oh I missed this! LGTM.

PR #408, opened 2016-08-15
New DB API Part #1 (See below)

erikh:
This new implementation is unfinished and contains several known issues (See
below). This implementation contains a new architecture which implements a
visitor model where the visitor (the Entity) fully contains the information to
be persisted to disk, where the processor (the Client) is responsible for
processing the object into a format which is capable to be written to the k/v
store.

This abstraction is necessary for Consul support.

Known issues (will be resolved in v2):

* Hooks calls return closures which are very expensive and generated each time
  Hooks is called.
* RuntimeOptions are not capable of being watched because the keyspace is not
  sent to the object, so the volume cannot be located.

**NOTE**: I'm calling executive veto on major design changes. This is the design. Let's focus on bugs and continuing improvement of it.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1,0 +1,75 @@
+package db
+
+/*
+
+client<->entity relationship:
+
+get/set/delete accept the Path() argument from entity to perform key/value
+lookup. Get and set respectively manage Read() and Write() from entity.
+*/
+
+// Client provides a method of hooking into our data stores. It processes
+// entities (see below) and their pipeline between them and k/v stores such as
+// etcd and consul.
+type Client interface {
+	// Get the entity. Will copy the contents of the JSON into the entity for
+	// use.
+	Get(Entity) error
+	// Set the entity. Will persist the entity to etcd based on the path it
+	// provides. (See Path())
+	Set(Entity) error
+	// Delete the entity from the database. Does nothing to the underlying
+	// Object.
+	Delete(Entity) error
+	// Watch the Entity; just the entity and nothing else. Returns a channel for
+	// Entity updates and one for errors.
+	Watch(Entity) (chan Entity, chan error) // watch of entity
+	// WatchStop stops a watch on a specific Entity.
+	WatchStop(Entity) error
+	// WatchAll watches the prefix of the entity for changes, watching all of the
+	// specific type.
+	WatchAll(Entity) (chan Entity, chan error) // watch of subtree
+	// WatchAllStop stops a WatchAll.
+	WatchAllStop(Entity) error
+	// Dump dumps a tarball make with mktemp() to the specified directory.
+	Dump(string) (string, error)
+	// Prefix returns the base prefix for our keyspace.
+	Prefix() string
+	// List takes an Entity which it will then populate an []Entity with a list of objects.
+	List(Entity) ([]Entity, error)
+}
+
+// Entity provides an abstraction on our types and how they are persisted to
+// k/v stores. They are capable of describing all they need to be written to
+// the database, which Client above is trained to do.
+type Entity interface {
+	// Prefix is the base path for all keys; used for watchall and list
+	Prefix() string
+	// Path to object; used for watch and i/o operations.
+	Path() (string, error)
+	// Validate the object and ensure it is safe to write, and to use after read.
+	Validate() error
+	// Copy the object and returns Entity. Typically used to work around
+	// interface polymorphism idioms in go.
+	Copy() Entity
+	// Hooks returns a db.Hooks which contains several functions for the Entity
+	// lifecycle, such as pre-set and post-get.
+	Hooks() *Hooks
+}
+
+// Hook is the type that represents a client hook in our entities system. Hooks
+// are fired at certain client operations (see Hooks below) to cleanup or
+// reconcile work.
+type Hook func(c Client, obj Entity) error
+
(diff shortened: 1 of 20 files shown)


db/db.go
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+// Client provides a method of hooking into our data stores. It processes
+// entities (see below) and their pipeline between them and k/v stores such as
+// etcd and consul.
+type Client interface {
+	// Get the entity. Will copy the contents of the JSON into the entity for
+	// use.
+	Get(Entity) error
+	// Set the entity. Will persist the entity to etcd based on the path it
```
vvb: @erikh: minor: since this is a generic, may be we can change "etcd" -> "kv store"

db/db.go
```diff
@@ -0,0 +1,75 @@
...
+	// provides. (See Path())
+	Set(Entity) error
```
vvb: @erikh question: are there any stores which differentiate Add and Set ? Add - first time, Set - change(could be a subset of the entry)? If so, we need to design for them here.

erikh: That’s a good question. Currently the usage patterns are enforced in 
code, which is necessary only for volumes at the moment.

The problem with “add vs set” however is really about what is adding 
and what is setting. In a distributed environment there could be two 
adds that happen at once or two sets. The semantics here get kind of 
wacky, but it could be done.

All that said until the usage patterns need it I would like to avoid 
amending this behavior.

db/db.go
```diff
@@ -0,0 +1,75 @@
...
+	// Object.
+	Delete(Entity) error
+	// Watch the Entity; just the entity and nothing else. Returns a channel for
+	// Entity updates and one for errors.
+	Watch(Entity) (chan Entity, chan error) // watch of entity
+	// WatchStop stops a watch on a specific Entity.
+	WatchStop(Entity) error
+	// WatchAll watches the prefix of the entity for changes, watching all of the
+	// specific type.
+	WatchAll(Entity) (chan Entity, chan error) // watch of subtree
```
vvb: WatchAll -> WatchHierarchy
I prefer it so, but really up to you.. does not bother me much..

erikh: How about WatchTree?

erikh: Actually let’s call this WatchPrefix as that’s literally what it 
does.

vvb: WatchPrefix is good.

dseevr:
Don't have anything further to add.  LGTM

db/impl/etcd/client.go
```diff
@@ -0,0 +1,291 @@
...
+		return errors.InvalidDBPath.Combine(errored.New("missing key during watch"))
+	}
+
+	close(stopChan)
+	delete(c.watchers, path)
+
+	return nil
+}
+
+func (c *Client) traverse(node *client.Node, obj db.Entity) []db.Entity {
```
yuva29: @erikh please explain me why do we need this func?

yuva29:
I don't have any comments. LGTM

db/global.go
```diff
@@ -0,0 +1,79 @@
...
+func NewGlobal() *Global {
+	return &Global{}
+}
+
+// Path returns the path to the global configuration
+func (g *Global) Path() (string, error) {
+	return rootGlobal, nil
+}
+
+// Prefix returns the path to the global configuration
```
vvb: "path" -> "prefix"

db/global.go
```diff
@@ -0,0 +1,79 @@
...
+	return nil
+}
+
+// Hooks returns an empty hooks set.
+func (g *Global) Hooks() *Hooks {
+	return &Hooks{}
+}
+
+// Copy returns a copy of this Global.
+func (g *Global) Copy() Entity {
```
vvb: @erikh Does the return type seem correct here?

I am trying to understand this,
```
type Global Struct {
    Debug     bool
    Timeout   time.Duration
    TTL       time.Duration
     MountPath string
}
```

Vs

```
type Entity interface {
Prefix() string
Path() (string, error)
Validate() error
Copy() Entity
Hooks() *Hooks
}
```

db/global.go
```diff
@@ -0,0 +1,79 @@
...
+	return rootGlobal, nil
+}
+
+// Prefix returns the path to the global configuration
+func (g *Global) Prefix() string {
+	return ""
+}
+
+// Validate validates the global configuration.
+func (g *Global) Validate() error {
```
vvb: "Validate" -> "Sanitize"

db/global.go
erikh: It may help you to learn a bit more about go’s interfaces: 
https://tour.golang.org/methods/4

db/global.go
erikh: The prefix is a path to itself; this is an appropriate word choice.

db/global.go
erikh: Sanitize would be better in this case but the interface prescribes validate. I'm not sure this is really the best place for it, but to ensure the values do get set this is the simplest choice.

Maybe moving this to a hook in v2 would be best.

db/impl/etcd/client.go
erikh: it traverses the keyspace and turns anything that looks like a passed entity into an entity to put into the list to return.

I'll put in a comment in v2.

PR #412, opened 2016-08-15
volmigrate: fix golint and misspell errors flagged by Go Report Card

dseevr:
Signed-off-by: Bill Robinson <dseevr@users.noreply.github.com>

PR #414, opened 2016-08-16
checks.sh: fix gofmt/golint/govet checks whose outputs were not being properly checked

dseevr:
gofmt sometimes includes blank lines in its output, and those are now stripped out.

Signed-off-by: Bill Robinson <dseevr@users.noreply.github.com>

### DIFF
--- a/build/scripts/checks.sh
+++ b/build/scripts/checks.sh
@@ -1,4 +1,4 @@
-#!/bin/sh	
+#!/bin/sh
 
 set -e
 
@@ -7,9 +7,10 @@ files=$(find . -type f -name '*.go' | grep -v ./vendor)
 
 echo "Running gofmt..."
 set +e
-out="$(gofmt -s -l ${files})"
+out=$(gofmt -s -l ${files})
 set -e
-if [ "$(echo ${out} | wc -l)" -gt 1 ]
+# gofmt can include empty lines in its output
+if [ "`echo \"${out}\" | sed '/^$/d' | wc -l`" -gt 0 ]
 then
   echo 1>&2 "gofmt errors in:"
   echo 1>&2 "${out}"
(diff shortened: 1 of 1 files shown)


dseevr:
Basically, our checks.sh isn't doing anything right now.  There's an issue where variables consisting of multiple lines of text being passed into `$( )` like `${foo}` are being collapsed into a single line before being passed to `wc -l`.

I have no idea if this is the right way to fix the escaping problem, but I was able to get it working using the last method here:

```
$ out=$(ls -l)
$ echo "$(echo ${out} | wc -l)"
1
$ echo "$(echo \"${out}\" | wc -l)"
1
$ echo "`echo \"${out}\" | wc -l`"
24
```

Even with these fixes so that line counts were properly obtained, the error output/exiting was not always firing.  If you had a `gofmt` error in one file, 1 is not `-gt 1` and it would never fire.  It would fire if you had errors in two separate files, etc.  I updated the checks to be 0 because of this.

`gofmt` sometimes includes pointless blank lines in its output so those are being stripped out now.

This CI build should fail because we currently have code in our repo (which we saw earlier on [Go Report Card](https://goreportcard.com/report/github.com/contiv/volplugin)) that was able to bypass these checks and make it through CI.  Once my PR (https://github.com/contiv/volplugin/pull/412) and @yuva29's pending PR are merged, I'll retrigger CI on this PR.

erikh:
# of yuva's pr?

erikh:
Thought I tested this pretty well but ok.

Thanks for fixing it.

I've merged #412, so let me know if this needs anything else.

dseevr:
Sorry, I should have said "forthcoming" not "pending".  She hasn't file a PR yet.

yuva29:
@erikh @dseevr Sorry, Here it is https://github.com/contiv/volplugin/pull/418

PR #415, opened 2016-08-16
build: ensure that new code runs in containers on make run

vvb:
When doing a make run, we need to clean up earlier contiv/volplugin docker
image from the system. This also mandates that any containers using the
image are stopped and removed.
Additionally, repeatedly running `docker build` can lead to dangling images.
These are also cleaned.

### DIFF
--- a/build/scripts/build-volplugin-containers.sh
+++ b/build/scripts/build-volplugin-containers.sh
@@ -1,4 +1,12 @@
 #!/bin/bash
+
+# In order for the new binaries to take effect we need to clean the older environment
+# Otherwise the binaries stick to older container images
+# repeated docker builds also lead to dangling images, which need to be cleaned too.
+docker ps -a | grep -e volplugin -e apiserver -e volsupervisor | awk '{print $1}' | xargs docker rm -fv
+docker rmi contiv/volplugin
+docker rmi $(docker images -f "dangling=true" -q)
+
 docker build -t contiv/volplugin .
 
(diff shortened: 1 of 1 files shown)


build/scripts/build-volplugin-containers.sh
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
+
+# In order for the new binaries to take effect we need to clean the older environment
+# Otherwise the binaries stick to older container images
+# repeated docker builds also lead to dangling images, which need to be cleaned too.
+docker ps -a | grep -e volplugin -e apiserver -e volsupervisor | awk '{print $1}' | xargs docker rm -fv
+docker rmi contiv/volplugin
+docker rmi $(docker images -f "dangling=true" -q)
```
erikh: I've read that using `dangling=true` can lead to some destructive behavior. Are you familiar with this at all?

vvb: yeah, I read up about this, so there are two type of dangling images, ones that show up when doing `docker ps -a` are the good ones and should not be removed. They are like the empty parents to existing good images. The ones that show up with the `-a` option are the bad ones. I have verified this locally on my setup multiple times.

erikh: awesome, thanks for the diligence.

erikh:
LGTM. Going to also merge this one on CI pass as it's likely breaking dev atm.

PR #419, opened 2016-08-16
Return volume at snapshot copy time

### LINKED ISSUE #413: Creating volume through snapshot - Not returning the new volume object
Volume-create is returning a new volume object after create,  Similarly Contiv-UI expects volume object to be returned after creating it using a snapshot. This would generalize the UI code across all functionalities.

erikh:
Fixes #413 

The snapshot copy api response was empty; the UI team would like it if it returned the volume that was created.

I have done this and also adjusted volcli to output the name of the new volume after copy.

### DIFF
--- a/apiserver/daemon.go
+++ b/apiserver/daemon.go
@@ -526,6 +526,18 @@ func (d *DaemonConfig) handleCopy(w http.ResponseWriter, r *http.Request) {
 		)).Combine(err))
 		return
 	}
+
+	content, err := json.Marshal(newVolConfig)
+	if err != nil {
+		api.RESTHTTPError(w, errors.PublishVolume.Combine(errored.Errorf(
+			"Creating new volume %q from volume %q, snapshot %q",
+			req.Options["target"],
+			volConfig.String(),
+			req.Options["snapshot"],
+		)).Combine(err))
+	}
+
+	w.Write(content)
 }
 
 func (d *DaemonConfig) handleGlobal(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


dseevr:
LGTM

apiserver/daemon.go
```diff
@@ -526,6 +526,18 @@ func (d *DaemonConfig) handleCopy(w http.ResponseWriter, r *http.Request) {
 		)).Combine(err))
 		return
 	}
+
+	content, err := json.Marshal(newVolConfig)
+	if err != nil {
+		api.RESTHTTPError(w, errors.PublishVolume.Combine(errored.Errorf(
```
yuva29: I'm not sure if we have to use some other error here instead `PublishVolume` because volume is successfully published L509.

erikh: The overall operation is publish; you'll notice almost all the other errors use this constant as well.

volcli/volcli.go
```diff
@@ -726,6 +727,18 @@ func volumeSnapshotCopy(ctx *cli.Context) (bool, error) {
...
+	if err != nil {
+		return false, errored.New("Reading body processing response").Combine(err)
+	}
+
+	vol := &config.Volume{}
+	if err := json.Unmarshal(content, vol); err != nil {
+		return false, errors.UnmarshalVolume.Combine(err)
+	}
+
+	fmt.Println(strings.Join([]string{vol.PolicyName, vol.VolumeName}, "/"))
```
yuva29: #413 issue states that they are expecting a volume object but here we return a string.

apiserver/daemon.go
yuva29: hmm.. okay.

volcli/volcli.go
erikh: Yes, this is unmarshalled data from the API server which we print in the CLI. The data itself came from the API server which they will query directly.

yuva29: cool.

yuva29:
LGTM

PR #424, opened 2016-08-17
db/volume.go: fixup new API according in the wake of c25c97e

erikh:
master is broken, this patch fixes it.

c25c97e9245784c7426b08c355e4a152d86415cc removed the backend.NewDriver call which this code relied on, and they both got merged today.


yuva29:
LGTM

erikh:
Unit tests passed and this is critical so merging.

PR #431, opened 2016-08-20
Makefile: wait for etcd on startup

erikh:
Will be quick-merging this if/when tests pass.

This (I think) fixes the issue with volcli global uploads not working on make restart or after etcd is terminated.


vvb:
LGTM

PR #435, opened 2016-08-22
Fix volsupervisor restart

erikh:
This corrects a number of issues with snapshot behavior in volsupervisor.

This is breaking most of the test runs right now so let's get this reviewed today please @yuva29 @dseevr.

### DIFF
--- a/lock/lock.go
+++ b/lock/lock.go
@@ -12,10 +12,8 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/coreos/etcd/client"
 	"github.com/jbeda/go-wait"
 
-	"github.com/contiv/errored"
 	"github.com/contiv/volplugin/config"
 	"github.com/contiv/volplugin/errors"
 )
@@ -108,11 +106,8 @@ func (d *Driver) ExecuteWithMultiUseLock(ucs []config.UseLocker, timeout time.Du
 // lock after a timeout, returning a stop channel. Timeout is jittered to
 // mitigate thundering herd problems.
 func (d *Driver) AcquireWithTTLRefresh(uc config.UseLocker, ttl, timeout time.Duration) (chan struct{}, error) {
+	// we acquire a permanent lock, then overwrite it with a TTL lock later.
 	if err := d.Config.PublishUse(uc); err != nil {
-		return nil, err
-	}
-
-	if err := d.acquire(uc, ttl, timeout); err != nil {
 		return nil, err
 	}
 
@@ -123,7 +118,7 @@ func (d *Driver) AcquireWithTTLRefresh(uc config.UseLocker, ttl, timeout time.Du
 			select {
 			case <-stopChan:
 				logrus.Debugf("Clearing lock for %v", uc)
-				if err := d.ClearLock(uc, timeout); err != nil {
+				if err := d.Config.RemoveUse(uc, false); err != nil {
 					logrus.Errorf("Could not clear lock %v after stop received: %v", uc, err)
 				}
 				return
@@ -173,15 +168,8 @@ func (d *Driver) acquire(uc config.UseLocker, ttl, timeout time.Duration) error 
 
 retry:
 	if ttl != time.Duration(0) {
-		if err = d.Config.PublishUseWithTTL(uc, ttl, client.PrevExist); err != nil {
+		if err = d.Config.PublishUseWithTTL(uc, ttl); err != nil {
 			logrus.Debugf("Lock publish failed for %q with error: %v. Continuing.", uc, err)
-			if er, ok := err.(*errored.Error); ok && er.Contains(errors.NotExists) {
-				if err = d.Config.PublishUseWithTTL(uc, ttl, client.PrevNoExist); err != nil {
-					logrus.Warnf("Could not acquire %q lock for %q: %v", uc.GetReason(), uc.GetVolume(), err)
-				}
-			} else if err != nil {
-				return err
-			}
 		}
 	} else {
 		if err = d.Config.PublishUse(uc); err != nil {
--- a/lock/lock.go
+++ b/lock/lock.go
@@ -130,7 +130,6 @@ func (d *Driver) AcquireWithTTLRefresh(uc config.UseLocker, ttl, timeout time.Du
 			case <-time.After(wait.Jitter(ttl/4, 0)):
 				if err := d.acquire(uc, ttl, timeout); err != nil {
 					logrus.Errorf("Could not acquire lock %v: %v", uc, err)
-					return
 				}
 			}
 		}
--- a/lock/lock.go
+++ b/lock/lock.go
@@ -108,19 +108,26 @@ func (d *Driver) ExecuteWithMultiUseLock(ucs []config.UseLocker, timeout time.Du
 // lock after a timeout, returning a stop channel. Timeout is jittered to
 // mitigate thundering herd problems.
 func (d *Driver) AcquireWithTTLRefresh(uc config.UseLocker, ttl, timeout time.Duration) (chan struct{}, error) {
+	if err := d.Config.PublishUse(uc); err != nil {
+		return nil, err
+	}
+
 	if err := d.acquire(uc, ttl, timeout); err != nil {
 		return nil, err
 	}
 
-	stopChan := make(chan struct{})
+	stopChan := make(chan struct{}, 1)
 
 	go func() {
 		for {
-			time.Sleep(wait.Jitter(ttl/4, 0))
 			select {
 			case <-stopChan:
+				logrus.Debugf("Clearing lock for %v", uc)
+				if err := d.ClearLock(uc, timeout); err != nil {
+					logrus.Errorf("Could not clear lock %v after stop received: %v", uc, err)
+				}
 				return
-			default:
+			case <-time.After(wait.Jitter(ttl/4, 0)):
 				if err := d.acquire(uc, ttl, timeout); err != nil {
 					logrus.Errorf("Could not acquire lock %v: %v", uc, err)
 					return
--- a/systemtests/util_test.go
+++ b/systemtests/util_test.go
@@ -27,8 +27,10 @@ var (
 )
 
 func ClearEtcd(node remotessh.TestbedNode) {
-	logrus.Infof("Clearing etcd data")
-	node.RunCommand(`etcdctl rm --recursive /volplugin`)
+	logrus.Info("Waiting for etcd")
+	node.RunCommand("while ! $(etcdctl cluster-health | tail -1 | grep -q 'cluster is healthy'); do sleep 1; done")
+	logrus.Info("Clearing etcd data")
+	node.RunCommand("etcdctl rm --recursive /volplugin")
 }
 
 // WaitForDone polls for checkDoneFn function to return true up until specified timeout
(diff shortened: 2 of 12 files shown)


volsupervisor/loop.go
```diff
@@ -16,6 +19,33 @@ var (
 	volumeMutex = &sync.Mutex{}
 )
 
+func (dc *DaemonConfig) updateVolumes() {
+	myVolumes, err := dc.Config.ListAllVolumes()
+	fmt.Println(myVolumes)
```
dseevr: Was this line of output for debugging?

erikh: oops

volsupervisor/loop.go
```diff
@@ -16,6 +19,33 @@ var (
...
+	if err == nil {
```
dseevr: We generally seem to avoid else clauses elsewhere in the codebase when possible.  Can we just return from a `err != nil` and then unindent the rest of the function?

volsupervisor/loop.go
```diff
@@ -16,6 +19,33 @@ var (
...
+		volumeMutex.Lock()
```
dseevr: `volumeMutex.Unlock()` can be deferred immediately here

volsupervisor/volsupervisor.go
```diff
@@ -42,8 +46,33 @@ retry:
...
+		logrus.Fatal("Could not start volsupervisor: already in use")
+	}
+
+	sigChan := make(chan os.Signal, 1)
+
+	go func() {
+		<-sigChan
+		logrus.Infof("Removing volsupervisor global lock; waiting %v for lock to clear", dc.Global.TTL)
+		stopChan <- struct{}{}
+		time.Sleep(dc.Global.TTL + 1*time.Second) // give us enough time to try to clear the lock
```
dseevr: IIRC, we moved away from using `1 * someDuration` in favor of just the bare duration.

```go
time.Sleep(dc.Global.TTL + time.Second)
```

volsupervisor/loop.go
erikh: sure.

volsupervisor/volsupervisor.go
```diff
@@ -42,8 +46,33 @@ retry:
...
+
+	signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
+
 	dc.signalSnapshot()
-	dc.updatePolicies()
+	dc.updateVolumes()
+	// doing it here ensures the goroutine is created when the first poll completes.
+	go func() {
+		for {
+			time.Sleep(1 * time.Second)
```
dseevr: Same here.  This can also be rewritten as

```go
for range time.Tick(time.Second) {
  // ...
}
```

Tick() sleeps before running the loop for the first time

volsupervisor/volsupervisor.go
erikh: yeah, derp.

lock/lock.go
```diff
@@ -108,19 +108,26 @@ func (d *Driver) ExecuteWithMultiUseLock(ucs []config.UseLocker, timeout time.Du
...
+	if err := d.Config.PublishUse(uc); err != nil {
+		return nil, err
+	}
+
 	if err := d.acquire(uc, ttl, timeout); err != nil {
 		return nil, err
 	}
 
-	stopChan := make(chan struct{})
+	stopChan := make(chan struct{}, 1)
```
dseevr: I understand the implication of this change (the first send on the channel will not block now), but what was the reason for the change?

volsupervisor/volsupervisor.go
erikh: I honestly like this form better, so unless you really object to it...

lock/lock.go
erikh: TTL refresh would hang in a number of spots otherwise while the sleep 
was happening.

systemtests/util_test.go
```diff
@@ -28,7 +28,7 @@ var (
 
 func ClearEtcd(node remotessh.TestbedNode) {
 	logrus.Infof("Clearing etcd data")
-	node.RunCommand(`etcdctl rm --recursive /volplugin`)
+	node.RunCommand(`connwait 127.0.0.1:2379 && etcdctl rm --recursive /volplugin`)
```
dseevr: Nice... the more sleeps/waits we can replace with connwait, the better off we'll be!

lock/lock.go
erikh: Here's a tweet:

https://twitter.com/erikhollensbe/status/767618650505261056

volsupervisor/loop.go
dseevr: Ahh, you later removed it in this commit anyways:  https://github.com/contiv/volplugin/pull/435/commits/1343d7c5366f85dc5e46909e9246681d7b157573

systemtests/util_test.go
erikh: indeed

volsupervisor/volsupervisor.go
dseevr: I don't care either way in this case

For loops which always sleep a fixed duration of time, I prefer `Tick()` because it communicates that's it's going to be sleeping right in the loop declaration rather than you seeing an infinite loop and having to look for a `Sleep()` somewhere in the body.  That's just personal preference, and I don't think it matters at all for tiny loops.

lock/lock.go
dseevr: Got it, thanks!

systemtests/volsupervisor_test.go
```diff
@@ -140,27 +115,24 @@ func (s *systemtestSuite) TestVolsupervisorRestart(c *C) {
 
 	volName := genRandomVolume()
 	fqVolName := fqVolume("policy1", volName)
-
 	c.Assert(s.createVolume("mon0", fqVolName, map[string]string{"unlocked": "true"}), IsNil)
 
-	time.Sleep(4 * time.Second)
```
dseevr: What was the reason for this big change in sleep time?  I'm having trouble seeing it from the diff

dseevr:
LGTM

systemtests/volsupervisor_test.go
erikh: Several changes were made to the systemtests.

Two tests weren't actually testing anything, that was one of them. 
(volumes now don't snapshot at all when the volume is mounted, but the 
test never mounted a volume so the snapshots were "working" but not 
correctly) Modifications were made to the policy to accommodate them, 
which cascaded into other changes.

erikh: not mounted. they snapshot when they're mounted now, and only then.

lock/lock.go
```diff
@@ -108,19 +108,26 @@ func (d *Driver) ExecuteWithMultiUseLock(ucs []config.UseLocker, timeout time.Du
 // lock after a timeout, returning a stop channel. Timeout is jittered to
 // mitigate thundering herd problems.
 func (d *Driver) AcquireWithTTLRefresh(uc config.UseLocker, ttl, timeout time.Duration) (chan struct{}, error) {
+	if err := d.Config.PublishUse(uc); err != nil {
+		return nil, err
+	}
+
 	if err := d.acquire(uc, ttl, timeout); err != nil {
```
yuva29: Don't we need to remove the use lock that's published at L111 ? because `acquire` might fail for lot of reasons. It need not be timeout always.

systemtests/volsupervisor_test.go
```diff
@@ -73,19 +48,19 @@ func (s *systemtestSuite) TestVolsupervisorSnapshotSchedule(c *C) {
 
 	c.Assert(s.createVolume("mon0", fqVolume("policy1", volName), map[string]string{"unlocked": "true"}), IsNil)
 
-	time.Sleep(4 * time.Second)
+	time.Sleep(10 * time.Second)
 
 	out, err := s.vagrant.GetNode("mon0").RunCommandWithOutput("sudo rbd snap ls policy1." + volName)
 	c.Assert(err, IsNil)
-	c.Assert(len(strings.Split(out, "\n")) > 2, Equals, true)
+	c.Assert(len(strings.Split(strings.TrimSpace(out), "\n"))-1 > 2, Equals, true)
```
yuva29: Is -1 for the `headers` that displayed during  L53?

lock/lock.go
erikh: it's overwritten during acquire

erikh: please provide an example (either code or description) which demonstrates the issues you are predicting/seeing.

systemtests/volsupervisor_test.go
erikh: yes

yuva29: Any specific reason why is it not been considered on other test cases ?

erikh: I have limited time. I can look into which ones are not accommodating 
the header, but I suspect all the ones that are (were) failing were 
related.

What we should really do is output these in the ceph json format and 
parse it that way, but alas, again, time.

volsupervisor/loop.go
```diff
@@ -16,6 +18,32 @@ var (
 	volumeMutex = &sync.Mutex{}
 )
 
+func (dc *DaemonConfig) updateVolumes() {
```
yuva29: ah. Earlier we never used to create snapshots for already existing volumes?

volsupervisor/loop.go
```diff
@@ -28,48 +56,48 @@ func (dc *DaemonConfig) pruneSnapshots(val *config.Volume) {
...
-		}
+	defer func() { stopChan <- struct{}{} }()
 
-		list, err := driver.ListSnapshots(driverOpts)
-		if err != nil {
-			logrus.Errorf("Could not list snapshots for volume %q: %v", val.VolumeName, err)
-			return err
-		}
+	driver, err := backend.NewSnapshotDriver(val.Backends.Snapshot)
+	if err != nil {
```
yuva29: I guess we need to clear the lock obtained at L59

volsupervisor/loop.go
```diff
@@ -28,48 +56,48 @@ func (dc *DaemonConfig) pruneSnapshots(val *config.Volume) {
...
 
-	if err != nil {
-		logrus.Errorf("Error removing snapshot for volume %q: %v", val, err)
+	for i := 0; i < toDeleteCount; i++ {
+		logrus.Infof("Removing snapshot %q for volume %q", list[i], val.VolumeName)
+		if err := driver.RemoveSnapshot(list[i], driverOpts); err != nil {
+			logrus.Errorf("Removing snapshot %q for volume %q failed: %v", list[i], val.VolumeName, err)
+		}
 	}
 }
```
yuva29: same comment as above.

erikh: read line 67 please, and the code in lock/lock.go for AcquireWithTTLRefresh. Pay specific attention to how `stopChan` is used.

yuva29: ah.. got it

yuva29:
LGTM

erikh:
the locking change I just made sent a ripple through the matrix, breaking everything. I actually am seeing error conditions here that should have been happening a long time ago (and now are), but I need to tweak some things.

I'm going to leave this open, if I can't close it in a few days, I will and return with a new PR later.

erikh:
let's try to do another review of this ASAP @dseevr @yuva29 if/when you get time. This is passing.

erikh:
sorry guys, this needs to be merged asap. review post-hoc and i will submit a new PR

PR #443, opened 2016-08-24
storage, config: Driver options as map[string]interface{}

yuva29:
This will let the DriverOptions to accept any data type. Helps new storage driver integrations.

### DIFF
--- a/storage/backend/ceph/ceph.go
+++ b/storage/backend/ceph/ceph.go
@@ -115,7 +115,12 @@ func (c *Driver) Create(do storage.DriverOptions) error {
 		return err
 	}
 
-	cmd := exec.Command("rbd", "create", mkpool(do.Volume.Params["pool"], intName), "--size", strconv.FormatUint(do.Volume.Size, 10))
+	poolName, err := do.Volume.Params.Get("pool", false)
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command("rbd", "create", mkpool(poolName, intName), "--size", strconv.FormatUint(do.Volume.Size, 10))
 	er, err := runWithTimeout(cmd, do.Timeout)
 
 	if er != nil {
@@ -150,7 +155,11 @@ func (c *Driver) Format(do storage.DriverOptions) error {
 
 // Destroy a volume.
 func (c *Driver) Destroy(do storage.DriverOptions) error {
-	poolName := do.Volume.Params["pool"]
+	poolName, err := do.Volume.Params.Get("pool", false)
+	if err != nil {
+		return err
+	}
+
 	intName, err := c.internalName(do.Volume.Name)
 	if err != nil {
 		return err
@@ -173,7 +182,10 @@ func (c *Driver) Destroy(do storage.DriverOptions) error {
 
 // List all volumes.
 func (c *Driver) List(lo storage.ListOptions) ([]storage.Volume, error) {
-	poolName := lo.Params["pool"]
+	poolName, err := lo.Params.Get("pool", false)
+	if err != nil {
+		return nil, err
+	}
 
 retry:
 	er, err := executor.NewCapture(exec.Command("rbd", "ls", poolName, "--format", "json")).Run(context.Background())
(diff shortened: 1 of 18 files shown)


erikh:
rebase please

erikh:
one more rebase, sorry

vvb:
LGTM

storage/backend/ceph/ceph.go
```diff
@@ -150,7 +150,7 @@ func (c *Driver) Format(do storage.DriverOptions) error {
 
 // Destroy a volume.
 func (c *Driver) Destroy(do storage.DriverOptions) error {
-	poolName := do.Volume.Params["pool"]
+	poolName := do.Volume.Params["pool"].(string)
```
erikh: if you do this and `do.Volume.Params["pool"]` is *not* a string, this will crash the program.

Please take it like so:

```go
poolName, ok := do.Volume.Params["pool"].(string)
if !ok {
  // generate an error
}
```

This will allow you to trap any conversion error.

Again, I can't express how important it is to not perform type assertions this way.

Please correct it in all the code.

yuva29: https://github.com/yuva29/volplugin/blob/036c561893e8b778b038eb92e78abda1b83b9d01/storage/backend/ceph/ceph.go#L557-L564 this validation ensures that its string or any expected type and this validation is called prior to any storage operations. Do we still need to capture the conversion error?

erikh: yes.

https://play.golang.org/p/2OTd_hvP76 run this.

storage/backend/ceph/ceph.go
```diff
@@ -115,7 +115,12 @@ func (c *Driver) Create(do storage.DriverOptions) error {
 		return err
 	}
 
-	cmd := exec.Command("rbd", "create", mkpool(do.Volume.Params["pool"], intName), "--size", strconv.FormatUint(do.Volume.Size, 10))
+	poolName, ok := do.Volume.Params["pool"].(string)
+	if !ok {
+		return errored.Errorf("Expected %q type for driver.pool", "string")
```
erikh: Let's not use the format string here since we're not using a variable

erikh:
LGTM other than the format string issue. Let me know when this has changed and we can merge.

yuva29:
I changed it. We can merge this after CI

storage/backend/ceph/util.go
```diff
@@ -13,7 +14,13 @@ func (c *Driver) MountPath(do storage.DriverOptions) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(c.mountpath, do.Volume.Params["pool"], volName), nil
+
+	poolName, ok := do.Volume.Params["pool"].(string)
+	if !ok {
+		return "", errored.Errorf("Expected string type for driver.pool")
```
erikh: this message should be a constant. we should not have it in many places.

sorry I did not realize this before.

erikh: Please put it in `errors/errors.go` in the appropriate area.

erikh: in fact, I think pulling out the string should just probably be its own function; this way the access is normalized and we can be a little safer that way. wdyt?

yuva29: Yeah. Please take a look now. I have added an another abstraction layer. Now, there is a generic `Get` for both nfs, ceph. And, I did not add any thing in `errors.go`; I feel its not required for the current design.  Let me know what you think.

storage/backend/ceph/ceph.go
```diff
@@ -554,8 +589,13 @@ func (c *Driver) Validate(do *storage.DriverOptions) error {
 		return err
 	}
 
-	if do.Volume.Params["pool"] == "" {
-		return errored.Errorf("Pool is missing in ceph storage driver.")
+	switch dVal := do.Volume.Params["pool"].(type) {
```
erikh: is this still required given the code above? couldn't you just check if `pool == ""`?

erikh:
LGTM after tests pass and my last comment is addressed

storage/backend/ceph/ceph.go
yuva29: Changed it. Hope it makes sense now.

storage/driver.go
```diff
@@ -146,3 +147,19 @@ func (v Volume) Validate() error {
 
 	return nil
 }
+
+// Get can be used to get only "string" types from element `driver`
+func (p Params) Get(attr string, allowEmpty bool) (string, error) {
```
erikh: so if I pass a nil instead of a string by accident and allowEmpty is set, it will be processed as a nil, not a string. Is this the behavior you expect?

yuva29:
Closing this. It requires more stable design. I'll reopen again once its ready.

PR #444, opened 2016-08-24
db: New Locking API

erikh:
This implements and adds a new interface called `Lock`. Lock is presented as an
extension to Entity; one must implement a few additional methods to support it.

A few new methods were also added to acquire and free locks both with TTLs and
persistent locks.

Tests were also added for the functionality.

### DIFF
--- a/db/impl/etcd/client.go
+++ b/db/impl/etcd/client.go
@@ -12,8 +12,11 @@ import (
 	"github.com/contiv/volplugin/db/jsonio"
 	"github.com/contiv/volplugin/errors"
 	"github.com/coreos/etcd/client"
+	wait "github.com/jbeda/go-wait"
 	"golang.org/x/net/context"
 )
+
+const etcdCodeNotFound = 105
 
 // Client implements the db.Client interface.
 type Client struct {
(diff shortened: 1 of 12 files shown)


db/impl/etcd/client.go
```diff
@@ -317,3 +318,141 @@ func (c *Client) ListPrefix(prefix string, obj db.Entity) ([]db.Entity, error) {
...
+		return errors.LockFailed.Combine(err)
+	}
+
+	path, err := lock.Path()
+	if err != nil {
+		return errors.LockFailed.Combine(err)
+	}
+
+	_, err = c.client.Set(context.Background(), c.qualified(path), string(content), &client.SetOptions{PrevExist: client.PrevNoExist})
+	if er, ok := err.(client.Error); ok && er.Code == 105 {
```
dseevr: What is 105 here?  Can we add a constant for this?

db/impl/etcd/client.go
```diff
@@ -317,3 +318,141 @@ func (c *Client) ListPrefix(prefix string, obj db.Entity) ([]db.Entity, error) {
...
+	}()
+
+	return stopChan, nil
+}
+
+// AcquireWithTTL acquires a lock with a TTL by using two compare-and-swap
+// operations to refresh any lock that exists by us, and to take any lock that
+// is not taken.
+func (c *Client) AcquireWithTTL(lock db.Lock, ttl time.Duration) error {
+	if ttl < 0 {
```
dseevr: Is there a use case for a 0 TTL lock?  I guess you could use it to check for the presence of a lock...

db/impl/etcd/client.go
```diff
@@ -317,3 +318,141 @@ func (c *Client) ListPrefix(prefix string, obj db.Entity) ([]db.Entity, error) {
...
+
+	opts := &client.SetOptions{
+		PrevExist: client.PrevExist,
+		TTL:       ttl,
+		PrevValue: string(content),
+	}
+
+	logrus.Debugf("Acquiring lock %v with ttl %v", lock, ttl)
+	_, err = c.client.Set(context.Background(), c.qualified(path), string(content), &client.SetOptions{TTL: ttl, PrevExist: client.PrevNoExist})
+	if er, ok := err.(client.Error); ok && er.Code == 105 {
```
dseevr: Here, too

dseevr:
LGTM

db/impl/etcd/client.go
erikh: The constant in our version of the library has the wrong code.

I'll comment this.

-Erik

db/impl/etcd/client.go
erikh: TTL 0 clears the TTL, but we should actually make a minimum; etcd gets 
really unreliable when you start writing under 250ms/hit.

erikh:
glad we haven't merged this yet, it's still breaking occasionally. I'm going to work on it.

erikh:
closing this, will revisit when consul is done

PR #452, opened 2016-09-01
Consul DB abstraction

erikh:
This implements consul support for the db subsystem; note that this is still
not integrated into the rest of volplugin.

Several changes to etcd were also made to unify interfaces.

Notable changes:

* Node abstraction to handle watch endpoints, metadata.
* Stop channels can now be closed (async) or sent a struct (sync) to abort a
  lock.
* New `helpers` library which provides generic wrappers for some CRUD and watch
  operations.
* Dump has been ported to use consul and abstracted away from etcd.
* New tests and many test fixes.


erikh:
@mapuri this was the ticket I was referring to on slack.

config/util_test.go
```diff
@@ -9,8 +9,21 @@ import (
 
 func stopStartEtcd(c *C, f func()) {
 	c.Assert(exec.Command("/bin/sh", "-c", "sudo systemctl stop etcd").Run(), IsNil)
-	time.Sleep(time.Second)
+	for {
+		if err := exec.Command("/bin/sh", "-c", "etcdctl cluster-health").Run(); err != nil {
+			break
+		}
+
+		time.Sleep(time.Second / 4)
```
unclejack: This should be `250 * time.Millisecond`.

config/util_test.go
```diff
@@ -9,8 +9,21 @@ import (
...
+
 	f()
+
 	c.Assert(exec.Command("/bin/sh", "-c", "sudo systemctl start etcd").Run(), IsNil)
-	time.Sleep(15 * time.Second)
+	for {
+		if err := exec.Command("/bin/sh", "-c", "etcdctl cluster-health").Run(); err == nil {
+			break
+		}
+		time.Sleep(time.Second / 4)
```
unclejack: This should be `250 * time.Millisecond` like above.

Makefile
```diff
@@ -33,6 +33,7 @@ clean:
 clean-vms:
 	@echo DO NOT USE THIS COMMAND UNLESS YOU ABSOLUTELY HAVE TO. PRESS CTRL-C NOW.
 	@sleep 20
+	pkill -9 VBoxHeadless
```
unclejack: Won't this nuke all running headless VirtualBox VMs, including those which don't belong to volplugin?

erikh: possibly!

erikh:
I'm fine with `time / 4`. I don't see any advantage over `250 * time.Millisecond`.  I presume all of us know enough math to figure that out.

erikh:
this needs a few touches (godeps specifically and some ineffassign changes) so it might be a bit before tests pass. Please review anyways since this is a large patch which I don't expect will change drastically.

Makefile
```diff
@@ -33,6 +33,7 @@ clean:
...
 	for i in $$(vboxmanage list vms | grep volplugin | awk '{ print $$2 }'); do vboxmanage controlvm "$$i" poweroff; vboxmanage unregistervm "$$i"; done
```
mapuri: unrelated to this patch, but I noticed that `unregister` command here is missing `--delete` flag, which will perhaps leave some remanants of the vm around.

db/impl/consul/client.go
```diff
@@ -0,0 +1,274 @@
...
+// Delete removes a key from consul.
+func (c *Client) Delete(obj db.Entity) error {
+	return helpers.WrapDelete(c, obj, func(path string) error {
+		_, err := c.client.KV().Delete(c.qualified(path), nil)
+		return err
+	})
+}
+
+// Watch watches stuff.
+func (c *Client) Watch(obj db.Entity) (chan db.Entity, chan error) {
```
mapuri: @erikh How does watch indicate the type of event on entity? In other words, does the the db.Client need to differentiate between create v/s modify v/s delete event on an Entity? May be there is no use case for this atm?

db/impl/consul/lock.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		return errors.LockFailed.Combine(errored.New("invalid lock requested to be freed (wrong host?)"))
+	}
+
+free:
+	select {
+	case <-mylock.monitorChan:
+	default:
+		mylock.lock.Unlock()
```
mapuri: This might be rare, but I think there might be a race here that can result in a panic, as the value pointed by `mylock` could potentially be changed (deleted) by another go routine trying to call Free()?

db/impl/consul/lock.go
```diff
@@ -0,0 +1,240 @@
...
+	c.lockMutex.Unlock()
+}
+
+// doAcquire does most of the work for the acquire* class of functions in the
+// consul client. See the other private functions in this file for more
+// information, but this is usually the top of them all.
+func (c *Client) doAcquire(obj db.Lock, ttl time.Duration, refresh bool) (chan struct{}, error) {
+	logrus.Debugf("Attempting to acquire %v", obj)
+
+	lockObj, running, err := c.lock(obj, ttl, refresh)
```
mapuri: This might again be rare but i think a race might exist here i.e. two routines attempting to lock same `obj` might both succeed in getting a new lock object of there own as the `c.locks` data structure get's updated much later below.

db/impl/consul/client.go
erikh: It does not. In fact, I may not even have a test for this, so let me circle back with you once I do have one. :)

db/impl/consul/lock.go
erikh: I don't think the re-pointing of mylock is possible with the current architecture; it's never reassigned after it meets the map, so this is quasi-FP with regard to value usage.

So there are two things happening here:

* the monitorchan closes when the lock expires. we don't need to do anything here.
* the alternative is to force the lock free.

Now, we should be able to rely on the consul client's behavior here -- to close the monitorchan if the lock is freed -- which will hit the first case in select and not attempt to double-free. The second access is protected by mutex (everywhere, I hope) by way of the value being gathered during the map access. This also ensures the a rewrite of that lock *to* the map does not affect the current value of this lock.

db/impl/consul/client.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// ListPrefix lists all the entities under prefix instead of listing the whole keyspace.
+func (c *Client) ListPrefix(prefix string, obj db.Entity) ([]db.Entity, error) {
+	pairs, _, err := c.client.KV().List(c.qualified(path.Join(obj.Prefix(), prefix)), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	entities := []db.Entity{}
```
yuva29: L231-L240 can be bundled as a function and reused in `List` func as well.

db/impl/consul/lock.go
erikh: We do a compare early in the `lock` call and eject the existing stopchan if anything matches (under mutex). I think this should work?

erikh: s/eject/return/

db/impl/consul/client.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+
+		entities = append(entities, copy)
+	}
+
+	return entities, nil
+}
+
+// ListPrefix lists all the entities under prefix instead of listing the whole keyspace.
+func (c *Client) ListPrefix(prefix string, obj db.Entity) ([]db.Entity, error) {
```
yuva29: How is `prefix` different than `obj.prefix`?

db/impl/consul/lock.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	logrus.Debugf("Attempting to free %q by %v", path, obj)
+
+	mylock, ok := c.getLock(path)
+	if !ok {
+		return errors.LockFailed.Combine(errored.New("Could not locate lock"))
+	}
+
+	if !reflect.DeepEqual(obj, mylock.obj) {
```
yuva29: Why do we need this check? L43 isn't it always `true`?

vendor/github.com/hashicorp/consul/api/session.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// Session returns a handle to the session endpoints
+func (c *Client) Session() *Session {
+	return &Session{c}
+}
+
+// CreateNoChecks is like Create but is used specifically to create
+// a session with no associated health checks.
+func (s *Session) CreateNoChecks(se *SessionEntry, q *WriteOptions) (string, *WriteMeta, error) {
```
yuva29: `CreateNoChecks` and `Create` can be unified using an extra flag.

yuva29: Please ignore. I didn't know that I'm reviewing vendor code 👀

db/impl/consul/client.go
```diff
@@ -0,0 +1,274 @@
...
+
+	return entities, nil
+}
+
+// WatchPrefixStop stops a watch.
+func (c *Client) WatchPrefixStop(obj db.Entity) error {
+	return helpers.WatchStop(c, obj.Prefix(), c.watchers, &c.watcherMutex)
+}
+
+// Dump dumps a tarball make with mktemp() to the specified directory.
```
dseevr: typo: make -> made

db/impl/etcd/client.go
```diff
@@ -175,86 +111,52 @@ func (c *Client) WatchStop(obj db.Entity) error {
...
+		if err != nil {
+			if err == context.Canceled {
+				logrus.Debugf("watch for %q canceled", wi.Path)
+				return
 			}
 
-			for _, entity := range c.traverse(resp.Node, obj) {
-				retChan <- entity
+			if _, ok := err.(*client.ClusterError); ok {
+				// silently return from the loop; this means the watch has died because
```
dseevr: Is this worth a `logrus.Debug()` ?  Could help anyone who hits this edge case

db/test/consul_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *testSuite) TestNewConsulClient(c *C) {
+	if !IsConsul() {
+		c.Skip("not consul")
+	}
+
+	client, err := getConsulClient()
+	c.Assert(err, IsNil)
+
+	volClient, err := consul.NewClient("volplugin", &api.Config{})
```
dseevr: Why are these three lines duplicated two other times in this test? (one with some better error formatting but identical otherwise)

db/impl/consul/lock.go
mapuri: > We do a compare early in the lock call

yes, that is correct. However, `lock()` will only return the stopchan if the underlying `c.locks` map has already been updated by one of the go-routines calling `doAcquire()`, right?

So if two go-rotuines were to call `doAcquire()` concurrently, there is a possibility that none of the routines have had a chance to update the `c.locks` map yet i.e. `lock()` method will create and return a new object for each routine. If this happens we will leak (and loose track of) one of the extra lock objects.

db/impl/consul/lock.go
erikh: the value won't be invalid, just the Unlock will fail. That's OK and should be expected, according to the consul API docs.

mapuri: ah, got it. Yes, if `mylock` value won't be invalid then this should be fine.

db/test/consul_test.go
erikh: yeah heh. I'll fix it.

Makefile
erikh: this would explain the out of space issues I see after running clean-vms a few times. incorporated, thanks!

db/impl/consul/client.go
erikh: `prefix` is passed in. `obj.Prefix()` is an interface method on `db.Entity` explained through `db/db.go`.

db/impl/consul/lock.go
erikh: no, it is not. Please read again.

db/impl/etcd/client.go
erikh: good idea.

erikh:
Ok, I'm pretty sure this is ready now.

@mapuri @yuva29 @unclejack @vvb @dseevr PTAL on tuesday.

db/impl/consul/lock.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if running {
+		return lockObj.doneChan, nil
+	}
+
+	if lockObj.refresh {
+		go c.doRefresh(lockObj)
```
mapuri: @erikh is this go routine expected to end by itself when a lock is freed by call to `Free()`? Somehow I didn't see any special handling in `doRefresh()` so was just wondering if we could possibly leak a go routine here.

mapuri:
LGTM. The reorg looks good. It's big rewrite though so I might have missed a few things :)

dseevr:
LGTM now that Madhav's questions have been answered

erikh:
will reboot CI in a bit (it should pass), thanks for reviewing folks

erikh:
also if anyone has comments now would be the time for them, once CI passes I will merge.

db/impl/consul/lock.go
erikh: yes, correct.

erikh:
merged with prejudice; we need to fix our CI.

robellison:
hi all, what is the status of this? is it going to be re-merged at some point?

PR #458, opened 2016-09-05
build/scripts/autorun-bootstrap.sh: restore policy and global upload into autorun container image

erikh:
The bug here is that if you use the autorun tools, the default policy and global weren't automatically uploaded. This does not line up with the documentation.


dseevr:
LGTM

PR #465, opened 2016-09-14
apiserver/daemon.go: fix bug where `volcli volume list $POLICY` showed all volumes, not just volumes belonging to the target policy.

### LINKED ISSUE #461: Volume list command logic fix
`volcli volume list <policy>` should display only the volumes of the given policy.

dseevr:
Fixes #461


erikh:
split out the test file into a separate commit; sorry, literally just fixed this in the consul port.

dseevr:
Done!

PR #475, opened 2017-02-22
Auto-remove alpine / test policy upload containers

andybrucenet:
Few containers are left around after the run...I just used --rm to dump them after they finish.


dseevr:
Thanks for your contributions!

